PR #2, opened 2011-11-26
Proxied models, delete() and PEP8

Kronuz:
I worked on these things:

```
Better support for delete() in the QuerySets (and a way to "globally" disable polymorphism)

Added proxied polymorphic models.

PEP 8 cleaning
```


adamn:
Seems like this PR should just be closed.

vdboor:
Agreed, these features have found their way into polymorphic already

PR #58, opened 2013-11-29
[Admin] Fully supports related fields (fk, m2m, 1to1, genericfk) to chil...

### LINKED ISSUE #46: Admin related lookup to a child model
Is there a way to achieve this without doing `admin.site.register(MyChildModel, MyParentModelAdmin)`?

BertrandBordage:
...d models.

This fixes #46, #47, and perhaps #57.
This also fixes some issues with grappelli related lookups.

WARNING: This changes the syntax of the admin integration.

There was no option but to change the syntax.  I tried keeping the same syntax and it was not working because while the admin site registry was iterated by Django, the size of the registry changed because of the lazy setup.  The only solution was to stick to the classical Django syntax.

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -21,7 +21,7 @@ The parent model
 The parent model needs to inherit ``PolymorphicParentModelAdmin``, and implement the following:
 
 * ``base_model`` should be set
-* ``child_models`` or ``get_child_models()`` should return a list
+* ``child_models`` or ``get_child_models()`` should return an iterable
   of Model classes.
 
 The exact implementation can depend on the way your module is structured.
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -14,7 +14,6 @@ The polymorphic admin is implemented via a parent admin that forwards the *edit*
 to the ``ModelAdmin`` of the derived child model. The *list* page is still implemented by the parent model admin.
 
 Both the parent model and child model need to have a ``ModelAdmin`` class.
-Only the ``ModelAdmin`` class of the parent/base model has to be registered in the Django admin site.
 
 The parent model
 ----------------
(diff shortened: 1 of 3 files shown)


sjdines:
@vdboor / @chrisglass is this still in the pipeline of interest? I had missed this PR and started working on a solution but this one looks good.

mrmachine:
I only briefly looked through the changes, but I have a couple of questions if this is going ahead.

1) Why do we need to return only models from `get_child_models()` and manually register the child models, instead of just automatically registering the child models?

2) If child models are registered with the admin, that means they will be visible in the admin's app/model dashboard? So there will be two ways to create and find/edit child objects, via the parent or directly via the child. This might be undesirable in some cases, and confusing to some end-users (although it is probably a requirement to support FKs to child models)?

sjdines:
@mrmachine 

1) You may not want all of the child models registered? I don't know the particular use case in practice though. Maybe instead of deleting old objects they could just be deactivated? This is all speculation on my part.

2) From the code I saw they would be shown. It is not a requirement to support FKs to child models (I have developed a way to do so but it has it's own issues such as have to modify `formfield_for_foreignkey`) but it may be the best way to go even if it could be more confusing.

mrmachine:
@sjdines

From reading the docs in the PR, it sounds like you *must* now register child models with the admin, as there is no longer any other way to link a `ModelAdmin` class to a child model. If that is the case, we could do it automatically, and people could always unregister and re-register if they needed to customise a 3rd party app that uses `django-polymorphic`?

Unless automatic registration cannot be done because it should happen at import time, but get_child_models() is called at runtime. Which brings up another point. One part of the configuration is now fixed (manual registration to admin at import time) and the other is dynamic (returning available child models). That might break for some people if they do not always return all child models in `get_child_models()`. I'm not sure if that is even a valid or common use case, but I suppose somebody might want to return only a subset of child models for a particular group of users?

LeGast00n:
For my part, I'm concerned about this new way of doing in the context of 
multiple model class inheritance, which is not common but still 
something I use.

vdboor:
thanks for the headsup :-)

PR #65, opened 2014-01-08
Feature: Create from super

JCoxwell:
I'm not sure if you want this feature, but I thought I'd offer it. Basically it allows you to convert a polymorphic model to a new class down the hierarchy (one level). For example, if ModelB subclasses ModelA then an instance of ModelA can become an instance of ModelB by adding the additional fields without changing it's primary key. 

I use this feature to create objects that need to be classified later.

There are 2 commits which can be applied separately:

One is the feature itself with a test. 

The other modifies the tests to work with MySQL. Apparently sqlite reuses ids where as MySQL does not, so I just push the ids into the repr strings for comparison. This is the simplest solution I could think of, but it's not ideal. At least they pass...


bckohan:
I have migrated this PR to #684

PR #95, opened 2014-08-18
Allow PolymorphicQuerySet.instance_of to accept strings.

ctbarna:
The interface is similar to passing strings into ForeignKey and helps resolve circular dependencies.

I can add tests and docs later but I wanted to get feedback on the approach first. I had originally put it in [`_create_model_filter_Q`](https://github.com/chrisglass/django_polymorphic/blob/master/polymorphic/query_translate.py#L213-L248) but you lose the "app label" awareness from the instance that allows you to pass in `ModelName` (as opposed to `app_label.ModelName`).


vdboor:
You can import model names locally in a function, so there is no need to use strings fortunately. :-)

PR #118, opened 2015-01-30
Set _default_manager directly on model_cls; don't use add_to_class()

### LINKED ISSUE #96: _default_manager cannot be accessed on a model instance
In non-polymorphic django models, it is possible to access `_default_manager` from a model instance but in polymorphic models, it raises `AttributeError: Manager isn't accessible via ModelName instances`.

fdintino:
Fixes #96 (_default_manager cannot be accessed on a model instance).

Using `model_cls.add_to_class()` calls `Manager.contribute_to_class()`, which then sets the attribute to an instance of `ManagerDescriptor`. In the case of `_default_manager`, this behavior differs from Django's, as `_default_manager` is set to the manager instance itself, not its descriptor. This is why it is normally possible to access `default_manager` from a model instance.


vdboor:
Thanks for taking care of this!

PR #119, opened 2015-01-31
Bug with queryset results on base model where child model has a custom primary key

### LINKED ISSUE #66: Custom primary key confusion
With a set of models like:
```
Owner(polymorphic), has many Projects
FooOwner(Owner), has custom primary key

Project(polymorphic), has many Owners
BarProject(Project), has custom primary key
```

performing `barproject.owners.add(FooOwner())` seemed to have no effect; afterwards, `barproject.owners.all()` would be empty.

When I went to write a test case to recreate, I noticed I was triggering [this line](https://github.com/chrisglass/django_polymorphic/blob/b73f22f4e0a98f1e0c65625b0293c65b533bfaef/polymorphic/tests.py#L377) in my test output. I then removed the custom primary keys from FooOwner and BarProject, which fixed the problem.

Is this a known issue? The closest mention I saw in the docs was about [unique primary keys in calls to extra()](https://django-polymorphic.readthedocs.org/en/latest/advanced.html#about-queryset-methods), but the test output seems to suggest somebody is aware of this.

fdintino:
There is currently a bug when inheriting from a PolymorphicModel with a custom primary key. For example, with the following models:

```python
class CustomPkBase(ShowFieldTypeAndContent, PolymorphicModel):
    b = models.CharField(max_length=1)

class CustomPkInherit(CustomPkBase):
    custom_id = models.AutoField(primary_key=True)
    i = models.CharField(max_length=1)
```

The following code results in a queryset that is missing items:

```python
CustomPkBase.objects.create(b='b')
CustomPkInherit.objects.create(b='b', i='i')
print(len(CustomPkBase.objects.all())) # expected: 2; actual value: 1
```

This pull request adds tests in the first commit (which fail), and then fixes those tests in the second commit. I believe that this pull request fixes #66.

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -154,6 +154,15 @@ class PolymorphicQuerySet(QuerySet):
         # in the format idlist_per_model[modelclass]=[list-of-object-ids]
         idlist_per_model = defaultdict(list)
 
+        # django's automatic ".pk" field does not always work correctly for
+        # custom fields in derived objects (unclear yet who to put the blame on).
+        # We get different type(o.pk) in this case.
+        # We work around this by using the real name of the field directly
+        # for accessing the primary key of the the derived objects.
+        # We might assume that self.model._meta.pk.name gives us the name of the primary key field,
+        # but it doesn't. Therefore we use polymorphic_primary_key_name, which we set up in base.py.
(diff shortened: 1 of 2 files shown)


vdboor:
Thanks for looking into this!

PR #120, opened 2015-01-31
Cleanup tests by removing spurious errors and unnecessary noise

fdintino:
- Remove diamond inheritance tests since these test a bug in Django,
  not in django_polymorphic.
- Fix spurious print message in test_primary_key_custom_field_problem(),
  where uuid instances are for some reason being compared to uuid.UUID
  classes. Replaced with a test that checks that the primary key values
  are valid UUIDs.
- Remove Django > 1.6 conditionals wrapping the multiple inheritance
  tests and model definitions, in favor of using the definitions for
  Django 1.7. Django 1.7 requires a custom primary key when using concrete
  model inheritance because failing to do so is potentially ambiguous. Adding
  the custom primary key does not break tests in older Django versions.

### DIFF
--- a/polymorphic/tests.py
+++ b/polymorphic/tests.py
@@ -67,45 +67,13 @@ class ModelX(Base):
 class ModelY(Base):
     field_y = models.CharField(max_length=10)
 
-if django.VERSION[:2] > (1, 6):
-    class Enhance_Plain(models.Model):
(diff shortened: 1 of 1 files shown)


vdboor:
Thanks for looking into this, this definitely improves future maintenance!

PR #124, opened 2015-03-02
Don't remove '/' from id when it's not in the path

benkonrath:
I get a 500 error when I try to load id without a trailing slash (i.e. /admin/app/parentmodel/id ). This fixes the issue.


vdboor:
Thanks Ben!

PR #126, opened 2015-03-10
replace long() with int() for py3 compatibility

skirsdeda:


vdboor:
Thanks for the pull request!
I've solved this issue in a slightly different way, so Python 2 code can still use long integers.

PR #127, opened 2015-03-11
Issue with related_name in polymorphic_ctype field

warrd:
Django 1.8b2 complains that the related name defined on polymorphic_ctype is not valid (due to this commit: django/django@1e5e2a470).

It requires either a valid Python identifier, or for the name to end in a '+'... so I added a '+' at the end.
Alternatively the period will have to be removed from the related name, and replaced with an underscore or something.

What do you think?


kraenhansen:
+1

vdboor:
Agreed! This identifier is purely needed to avoid duplicate related names. It serves no other purpose. And clearly, because it's an invalid identifier, I don't expect it to be used.

PR #130, opened 2015-04-06
fixed Django 1.8 complain about related name

tyaslab:
Django 1.8 complains because of invalid regex. So, I tried to to remove . (dot) character and it worked.


vdboor:
Hi! Thanks for the PR!

I've merged the source_expressions commit.
The related_name issue was addressed in a better way in #127, I've merged that PR instead.

PR #139, opened 2015-06-08
Use parent admin site class for child admin site

martinmaillard:
Using the `AdminSite` class directly can cause issues when the parent
model is registered with a custom admin site class.

For example in my project I have a custom `User` model that doesn't have an `is_staff` flag. Since the `AdminSite.has_permission` method uses this flag, I had to override it in a custom admin site class. Now when `real_admin.change_view` is called, an error is raised because the original `AdminSite.has_permission` is executed instead of my custom method.


vdboor:
Thanks, good fix!

PR #149, opened 2015-07-12
add base_model declaration to child models in admin docs example (docfix)

kelseyq:
Updates the example code in the admin section of the docs to reflect that "base_model should be set in [the child model admin] class as well." For those of us who don't read unless absolutely forced to...


vdboor:
Thanks for the PR!

PR #159, opened 2015-09-21
Raise Http404 if object does not exist.

mrmachine:
When visiting the change form for a polymorphic model and providing a bogus primary key, we currently get a 500 error `ObjectDoesNotExist`. This change raises `Http404` instead.


vdboor:
Thanks!

PR #161, opened 2015-09-25
Add support for qset.only() and qset.defer()

fdintino:


fdintino:
Hmm, this seems to fail on django 1.4. I'll take a look.

fdintino:
It looks like there [was a bug](https://code.djangoproject.com/ticket/14694) that prevented `defer()` from working with reverse relations that wasn't [fixed](https://github.com/django/django/commit/a21e8dee7682abdba646c5b72b6371d392e44dd1) until 1.5. My inclination is to wrap that one part of the test in `if django.VERSION > (1, 5)` and have this limitation noted in the docs.

Speaking of which, would you like me to update the docs as part of this pull request?

vdboor:
Wow, this is a great improvement! Thanks for the work on this, and thanks for providing proper tests too.

Yes, updating the docs would be a good addition to this PR!
A 0.7.2 release will happen soon, but since this feature could be in 0.8.

polymorphic/tests.py
```diff
@@ -395,6 +395,72 @@ def test_simple_inheritance(self):
...
+            '<Model2C_Deferred_field2_field3_id_model2a_ptr_id: '
+            'id 3, field1 (CharField), field2 (CharField), field3 (CharField)>')
+        self.assertEqual(repr(objects_only[3]),
+            '<Model2D_Deferred_field2_field3_field4_id_model2a_ptr_id_model2b_ptr_id: '
+            'id 4, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField)>')
+
+        # A bug in Django 1.4 prevents using defer across reverse relations
+        # <https://code.djangoproject.com/ticket/14694>. Since polymorphic
+        # uses reverse relations to traverse down model inheritance, deferring
+        # fields in child models will not work in Django 1.4.
```
vdboor: Suggestion: can you place this part in a separate unit test like:

 ```
@skipIf(django.VERSION < (1,5))
def test_defer_relation_fields()
```

That makes it more explicit that this feature is only for 1.5+

fdintino:
I've updated the docs in the PR.

fdintino:
I'm disappointed. I was hoping this would be included in the 0.8 release.

vdboor:
I'm sorry, I've missed this PR in the rush to get a Django 1.9 compatible release out. I've merged this change now!

PR #170, opened 2015-10-22
Django19

### LINKED ISSUE #164: request.REQUEST deprecated, removed in Django 1.9
```python
/lib/python3.4/site-packages/polymorphic/admin.py:383: RemovedInDjango19Warning: `request.REQUEST` is deprecated, use `request.GET` or `request.POST` instead.
  'is_popup': "_popup" in request.REQUEST,
```

### LINKED ISSUE #166: Django 1.9 support
Django 1.9 support probably needs multiple small changes here and there, but one of the first issues to be solved is how to deal with Django 1.9's new app loading behaviour. Since Django 1.8, loading models directly or indirectly from an app's `__init__.py` is deprecated. As of Django 1.9, it's not working anymore at all. See this thread: https://groups.google.com/forum/#!topic/django-developers/26GdwpmU6KQ

The `shortcuts` workaround seems like a good idea for me.

I'm happy to come up with a patch.

hobarrera:
This adds support for django 1.9, as well as dropping some other deprecated features.

Closes #166.
Closes #164.

This changeset has backwards-incompatible changes. I suggest a pushing the major in next release after merging this.

Noticably  import path for models, managers, etc has been moved (both to stay in line with django standards, and to avoid crashes).


hobarrera:
There are some issues with the admin (a redirection one with the test app, and another which I can only repro on a local project).

Feel free to poke around though.

jonashaag:
Do we need the migration file?

jonashaag:
I made a few adjustments here: https://github.com/jonashaag/django_polymorphic/commit/e09a59b69c73aaed36cfe225108a09cc3559cf35

hobarrera:
> Do we need the migration file?

Without it, we can't create a clean database to run the example file, so, yes. There's no harm in it being present for older versions of django anyway.

jonashaag:
Oops, this breaks the build for Django < 1.9, I'll fix that tomorrow.

jonashaag:
Should work now.

hobarrera:
Added them to the PR, thanks. There's still an admin redirection issue (as seen in the example app), which I'll try to look at during the weekend.

hobarrera:
I just added an example (via 4f0a9de) for a scenario that reproduces the last pending issue:

* Create a `ModelC` instance.
* Relate it to one or more `ModelB`s.
* Open the change page in the admin.

Crash.

I can't quite wrap my head around this one, so I'll appreciate any help I can get.

jonashaag:
> I just added an example (via 4f0a9de) for a scenario that reproduces the last pending issue:

There doesn't seem to be a unit test covering this case. Is that correct? How did you discover the issue then?

hobarrera:
> There doesn't seem to be a unit test covering this case. Is that correct? How did you discover the issue then?

By attempting to port an existing app to django1.9. Using it for a while and triggered this issue pretty fast (since I have two models with basically that same relationship in it). I left a big TODO to add a unittest to cover that at some point for that exact reason.

jonashaag:
Note: The migrations don't work on Django < 1.8.

jonashaag:
I added a fix in my branch, please pull.

hobarrera:
@jonashaag Thanks for that last fix. You clearly have a better understand of both django_polymorphic and django's `QuerySet` than I do. 😁

I'd consider this no longer WIP.

jonashaag:
> @jonashaag Thanks for that last fix. You clearly have a better understand of both django_polymorphic and django's QuerySet than I do. 😁

Not really; at best you can say I looked at the right places. I wondered why the super `iterator` returned integers on Django 1.9 (whereas it returns objects on Django < 1.9), so I compared the `iterator` implementations of Django 1.8 and 1.9 and found the behaviour change.

hobarrera:
> so I compared the `iterator` implementations of Django 1.8 and 1.9 and found the behaviour change.

I wonder why I didn't think of that.

Well, this is ready to merge IMHO. I'll include another PR improving test case running and adding some missing scenarios (I'd like to do quite a bit of cleanup there separately).

hobarrera:
bump?

hobarrera:
bump again? Any news?

hobarrera:
Hello? Any feedback on this?

noxan:
+1
keeps me from upgrading to 1.9 and pull request seems to work perfectly so far.

hobarrera:
It's been about two months since the last activity by the author, so I've made a fork of this, with dj19 support.  
Hopefully we can have that merged back in here at some point, but since I need an upstream package to depend on for a project of mine, I've no choice for now:

**django-polymorphic-ng**:
 * [github](https://github.com/hobarrera/django-polymorphic-ng)
 * [pypi](https://pypi.python.org/pypi?:action=display&name=django-polymorphic-ng&version=0.8.0)

macolo:
@hobarrera thanks for this!

@chrisglass ??

polymorphic/admin.py
```diff
@@ -465,7 +477,8 @@ def get_form(self, request, obj=None, **kwargs):
         kwargs.setdefault('form', self.base_form or self.form)
 
         # prevent infinite recursion in django 1.6+
-        if not self.declared_fieldsets:
+        if not hasattr(self, 'declared_fieldset') or \
```
vdboor: I guess this is a typoo, and should be `declared_fieldsets` with an s ?
It could be replaced with `if getattr(self, 'declared_fieldsets', None):` and catch the exists check and bool check in one go.

polymorphic/query_translate.py
```diff
@@ -162,7 +162,7 @@ def translate_polymorphic_field_path(queryset_model, field_path):
         # so no tripple ClassName___field was intended.
         try:
             # rel = (field_object, model, direct, m2m)
-            field = queryset_model._meta.get_field_by_name(classname)[0]
+            field = queryset_model._meta.get_field(classname)
```
vdboor: I think this breaks Django 1.7, right?

polymorphic/tests.py
```diff
@@ -15,9 +19,14 @@
...
-from polymorphic import PolymorphicModel, PolymorphicManager, PolymorphicQuerySet
+from polymorphic.models import PolymorphicModel
+from polymorphic.manager import PolymorphicManager
+from polymorphic.query import PolymorphicQuerySet
 from polymorphic import ShowFieldContent, ShowFieldType, ShowFieldTypeAndContent
-from polymorphic.tools_for_tests import UUIDField
+if django.VERSION < (1,8):
+    from polymorphic.tools_for_tests import UUIDField
+else:
+    from django.db.models import UUIDField
```
vdboor: Perhaps handle this with an try..except too, like the other code? (and comment that it's for django 1.9+)

vdboor:
Thanks for the great work on this pull request! Sorry for not getting back to this earlier.
I've added some comments that need to be looked at. When that's done, I'm happy to merge it!

polymorphic/query_translate.py
jonashaag: Nope, `get_field` has there since Django 1.0.

polymorphic/tests.py
jonashaag: I disagree. `if version ...` makes it very clear that this was necessary in Django < 1.8 and isn't necessary anymore for Django >= 1.8.

hobarrera: Sure, that's a lot more pythonic. Done (note that it's for 1.8+).

polymorphic/query_translate.py
hobarrera: No, at least tests don't reflect it. Doc doesn't say that this is new to 1.8 either.

polymorphic/admin.py
hobarrera: Well thought, fixed.

hobarrera:
I've fixed everything mentioned in the comments (note that I've rewritten history to keep this PR tidy).

hobarrera:
Thanks!

I've a few more changes on my fork; I'll send PRs for those ASAP and then get rid of it once we have a new release. :)

polymorphic/tests.py
jonashaag: Please see my comment. I think it's stupid to use `try... except` here.

polymorphic/query_translate.py
jonashaag: It's been undocumented but stable API. Have a look at the code.

polymorphic/tests.py
jonashaag: Also it's now inconsistent with https://github.com/chrisglass/django_polymorphic/blob/a01fd92d1c2a65f06dc1f5e4b0e96e459b0d59d8/example/pexp/models.py#L38

polymorphic/query_translate.py
hobarrera: Great, so I guess it's fine to keep this as-is.

vdboor:
Thanks for your work on this!

I noticed a few issues with this change, perhaps you can look into those?
* The support for fields starting with an underscore broke. so a query of `translations___field` (3 underscores), is recognized as `translations` class, whereas the old code also tried a join on `translations._field` as second option if that didn't work.
* The removed imports from `__init__.py` break other packages, as `from polymorphic.models import ..` does not work with older versions. They have to implement an `try.. import except` flow now. Would it make sense to keep the imports in `__init__.py` for Django <= 1.8 ?

hobarrera:
> The support for fields starting with an underscore broke. so a query of translations___field (3 underscores), is recognized as translations class, whereas the old code also tried a join on translations._field as second option if that didn't work.

I'll try and have a look at that, I'm not too familiar with that feature (yet?).

> The removed imports from __init__.py break other packages, as from polymorphic.models import .. does not work with older versions. They have to implement an try.. import except flow now. Would it make sense to keep the imports in __init__.py for Django < 1.9 ?

I don't really see the sense of keeping the imports, maybe documenting the API change for the latest release would be best. Note that this even fired a warning on django 1.8, since models were imported before app initialization, so even if it works, it's still not the recommended way.

PR #176, opened 2015-12-16
PEP-8 compliance

hobarrera:
Make all the code pep8 and flake8 compliant. Also add tests to make sure it stays tidy.

### DIFF
--- a/example/example/settings.py
+++ b/example/example/settings.py
@@ -67,15 +67,16 @@ INSTALLED_APPS = (
     'django.contrib.admin',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
-    #'django.contrib.sites',
+    # 'django.contrib.sites',
     'django.contrib.messages',
     'django.contrib.staticfiles',
 
-    'polymorphic',      # needed if you want to use the polymorphic admin
-    'pexp',             # this Django app is for testing and experimentation; not needed otherwise
+    'polymorphic',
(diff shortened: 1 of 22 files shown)


vdboor:
This might not be a polular opinion, but I do prefer something like 120 line size for readability (open for suggestions). The white space fixes, I'd like to have fixed indeed.

In one of my other projects, I found it safe to:
Update on: E112,E113,E115,E116,E225,E226,E227,E228,E231,E271,E272,E273,E274,E301,E302,E303,E304,E309,W291,W293,W391

would that be ok for a start? Line size can be discussed later, I find it cluttering the commit a lot.

hobarrera:
> This might not be a polular opinion, but I do prefer something like 120 line size for readability (open for suggestions).

There are too issues with this:
 * Most contributors will have to configure their editor/IDE specifically for this project or get a bunch of warnings.
 * Doing side-by-side diffs becomes almost impossible for anyone using a readable font size (including myself).

vdboor:
With the exception for line-size, everything else is now PEP8 compliant. I'll review the line-size issue later.

hobarrera:
The output of flake8 also reveals:

 * unused imports
 * never-read variables
 * whitespace issues
 * membership testing via `not a in b` instead of `a not in b`

I honestly don't see much reason to not-merge this right now, since it bears no functional changes or downsides.

vdboor:
Most flake8 issues are fixed now too.

PR #177, opened 2015-12-16
SCM versioning

hobarrera:
Use setuptools_scm for versioning.

setuptools_scm determines the package version on buildtime (or install time) using the latest git tag. So if a clean checkout of "1.0.0" is done, the wheel automatically has that number. It also tags dirty build appropriately.

All this:

 * Reduces noise avoiding "push version" commits in project's history.
 * Makes sure tags and versions are consistent everywhere.
 * Makes dirty builds easily recognizable.


hobarrera:
Huh. This doesn't seem to work with py32.  
Maybe we can leave this pending until py32 support is dropped.

vdboor:
I'd rather keep things simple here. The version commit isn't noise in my opinion, it clearly marks the boundary between two versions and everyone can learn from it to know how a release is done.

hobarrera:
A tag honestly also marks a clear boundry, and I don't see what there is to learn; there's nothing special "done" about pushing a new release (it can also be automated on CI for each tag, BTW).

Issue #64, opened 2013-12-18
_default_manager can't be PolymorphicManager when adding user_managers


kapt:
Hi,

If we define a new custom manager on a model as defined in the doc: https://django-polymorphic.readthedocs.org/en/latest/managers.html#using-a-custom-manager, it is not possible to keep the PolymorphicManager() as _default_manager because the classmethod get_first_user_defined_manager() skip the manager if its type is PolymorphicManager:

for key, val in new_class.**dict**.items():
            if isinstance(val, ManagerDescriptor):
                val = val.manager
            if not isinstance(val, PolymorphicManager) or type(val) is PolymorphicManager:
                continue

According to the following example of the doc: 
class Project(PolymorphicModel):
     objects = PolymorphicManager()               # add the default polymorphic manager first
     objects_ordered = TimeOrderedManager()       # then add your own manager
     start_date = DateTimeField()                 # project start is this date/time

It means that _default_manager will be filled with TimeOrderedManager().
Is it normal ? 

If the  Project class doesn't redefined objects = PolymorphicManager(), I agree with this behavior, but if objects = PolymorphicManager() is redefined in the class, I think that it means 'use PolymorphicManager as _default_manager'.

gilgamezh:
bump

bckohan:
This was closed by #186

PR #186, opened 2015-12-24
Allow to set a 'extra' custom manager withut override the default

gilgamezh:
from https://docs.djangoproject.com/en/1.8/topics/db/managers/

>Default managers
>
>If you use custom Manager objects, take note that the first Manager Django encounters (in the order >in which they’re defined in the model) has a special status. Django interprets the first Manager >defined in a class as the “default” Manager, and several parts of Django (including dumpdata) will >use that Manager exclusively for that model. As a result, it’s a good idea to be careful in your choice >of default manager in order to avoid a situation where overriding get_queryset() results in an inability >to retrieve objects you’d like to work with.

Same here --> https://django-polymorphic.readthedocs.org/en/latest/managers.html#custom-managers-querysets-manager-inheritance

So if a manager named `objects` is set before a custom manager. the `objects` manager should be the `_default_manager`

related to #64


vdboor:
thanks for the clear pull request! Makes it easy to merge!

gilgamezh:
thanks to you! :smile:

gilgamezh:
@vdboor do you have an ETA for the next release?

PR #188, opened 2015-12-29
Bug in example code

retailify:
"from polymorphic.model import PolymorphicModel" should
"from polymorphic import PolymorphicModel"


vdboor:
We changed the imports for the 0.8 release (to handle Django 1.9), so this is in fact correct.

Issue #81, opened 2014-05-15
UUID field and admin error


matteofattore-nextage:
My models use UUID as primary-key. I used cramer/django-uuidfield, because I got a problem using south, if I use UUID field defined in project test_tools.

If I use django_polymorphic and cramer/django-uuidfield together, I get this error, when I try to show a models list in admin interface:

```
Reverse for 'base_devices_basedevice_change' with arguments 
'(UUID('9639a9909ed24a99ae20f2c0f5c631ba'),)' and keyword arguments '{}' not found. 
1 pattern(s) tried: [u'admin/base_devices/basedevice/(\\d+)/$']
```

I solved it changing in PolymorphicParentModelAdmin class, line 250:
```python
new_change_url = url(r'^(\d+)/$', self.admin_site.admin_view(self.change_view), name='{0}_{1}_change'.format(*info))
```
to
```python
new_change_url = url(r'^(\w+)/$', self.admin_site.admin_view(self.change_view), name='{0}_{1}_change'.format(*info))
```

Have somebody a better solution??

zenzire:
Issue can be closed.
Adding pk_regex variable to Admin class fixes the problem with UUID primary key in admin panel.

```
class AttributeAdmin(PolymorphicParentModelAdmin):
    pk_regex = '([\w-]+)'
    base_model = Attribute
```

vdboor:
Thanks for providing the solution @zenzire, closing the issue.

PR #190, opened 2015-12-30
Mention pk_regex in docs when using non-int pk

amuttsch:
I stumbled upon this and figured it out by checking the issue tracker. It should be mentioned in the docs that non-int pks require additional work to work properly.


amuttsch:
See #81

vdboor:
Thanks!

PR #194, opened 2016-01-21
Support for Django 1.8 complex expressions in annotate/aggragte

slide333333:
This fixes [171](https://github.com/chrisglass/django_polymorphic/issues/171).
It allows using complex expressions in aggregate/annotate on a PolaymorhicQueryset like:
```
MyModel.poly_manager.aggregate(cnt=Count(Case(When(some_field=1, then=1))))
```


vdboor:
Wonderful, thanks!

PR #196, opened 2016-01-25
Update quickstart for Django 1.9 compatibility

GregoryAveryWeir:
The changelog states that 1.9 is supported, but the quickstart document has not yet been updated.


vdboor:
Thanks!

PR #206, opened 2016-04-01
Re-add convenience imports back in using some module magic

fdintino:
The convenience imports in `polymorphic/__init__.py` were, for us, extremely convenient. There is a way to lazily allow these imports rather than automatically importing them on load, which gets around the issues with how Django 1.9+ requires a stricter order of imports. The principle behind it is stolen from [werkzeug](https://github.com/pallets/werkzeug/blob/41177bf0efc198383ce9e692ec62be8e5fb7fcff/werkzeug/__init__.py), and the hack of swapping in a class instance in `sys.modules` is [Guido approved](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html).

### DIFF
--- a/polymorphic/__init__.py
+++ b/polymorphic/__init__.py
@@ -10,6 +10,9 @@ Please see LICENSE and AUTHORS for more information.
 __version__ = "0.9.1"
 
 
+import sys
+from types import ModuleType
+
 # Monkey-patch Django < 1.5 to allow ContentTypes for proxy models.
 import django
 if django.VERSION[:2] < (1, 5):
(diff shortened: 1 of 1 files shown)


vdboor:
hmm, interesting approach! :-) I do wonder, this isn't really "one way to do it". I managed to port my projects rather quickly, but I understand that's a bigger issue in your situation?

fdintino:
No, it's not a bigger issue. I could feign laziness, but it certainly would have been less effort to change the imports than to make this pull request :-)  I wouldn't necessarily disagree with you if you felt that this "fix" is more trouble than it's worth. Chacun à son goût and what not.

vdboor:
I'm sorry, I'd rather opt for the choice of having "one way to do it" (tm). Much time have passed too, so it would make sense most projects have cleaned up their imports by now.

Issue #201, opened 2016-02-26
AttributeError: 'NoneType' object has no attribute 'name'


jhg:
```python
  File "/opt/Dev_env/lib/python3.4/site-packages/django/contrib/admin/templatetags/admin_list.py", line 378, in date_hierarchy
    months = getattr(months, dates_or_datetimes)(field_name, 'month')
  File "/opt/Dev_env/lib/python3.4/site-packages/django/db/models/query.py", line 650, in datetimes
    plain_field=F(field_name)
  File "/opt/Dev_env/lib/python3.4/site-packages/polymorphic/query.py", line 238, in annotate
    self._process_aggregate_args(args, kwargs)
  File "/opt/Dev_env/lib/python3.4/site-packages/polymorphic/query.py", line 233, in _process_aggregate_args
    patch_lookup(a)
  File "/opt/Dev_env/lib/python3.4/site-packages/polymorphic/query.py", line 197, in patch_lookup_gte_18
    patch_lookup_gte_18(source_expression)
  File "/opt/Dev_env/lib/python3.4/site-packages/polymorphic/query.py", line 199, in patch_lookup_gte_18
    a.name = translate_polymorphic_field_path(self.model, a.name)
AttributeError: 'NoneType' object has no attribute 'name'
```

vdboor:
I've fixed the issue, but it needs some unit testing to clarify why things happened.

jhg-tgu:
I have in admin a model that in version 0.8.1 haven't this error but when upgrade to 0.9.0 or 0.9.1 raise this error in same line of polymorphic.

vdboor:
I'm sorry, it seems this is still only fixed in the master branch (5aa15b226e0215ac98c3f9455bb0ac4f2f6be610). Do you have some code examples that trigger this issue? I'd like to have this secured by tests before I make the release.

NotSqrt:
Hi,

Just faced this issue, I have isolated the problem to the use of Expressions before founding this issue. Here is a minimal case that should reproduce the issue:

```python
class MyModel(PolymorphicModel):
    date = DateTimeField()

from django.db.models.expressions import DateTime
MyModel.objects.all() # OK
MyModel.objects.all().annotate(day_round=DateTime('date', 'day', pytz.utc))  # < fails because day_round.get_source_expressions() == [None]
```

vdboor:
Released in 0.9.2!

PR #208, opened 2016-04-26
Test that expressions are supported

NotSqrt:
Test for #201 

I checked that without commit 88922d7d5e918f1338eeb9baeecb17ed98ed1261, the test fails.

I'm waiting for travis to see if it works on all django versions.


vdboor:
Thanks for the test! This also closes #201 then

Issue #200, opened 2016-02-19
Including super-types of proxy models


andrewdodd:
Hi,

I am not sure if this is an issue or not. I have investigated if there are existing open or closed issues about this and have searched the docs to see if this is 'desired' behaviour, but I can't find an answer.

The issue I am seeing is that the querysets returned by 'proxy' sub-types includes objects from the super type.

For example, if I have the following models:

```
class Drink(PolymorphicModel):
    name = models.CharField(max_length=50)
    colour = models.CharField(max_length=50)

    def __str__(self):
        return self.name


class AlcoholicDrink(Drink):
    strength = models.FloatField()
    

class HotDrink(Drink):
    class Meta:
        proxy = True

    def __str__(self):
        return '{}, really gets you going!'.format(self.name)
```

And I create the following objects:

- Drinks => Water, Coke
- HotDrinks => Coffee
- AlcoholicDrinks => Beer

Via shell:
```
>>> Drink.objects.create(name='Water', colour='Clear')
<Drink: Water>

>>> Drink.objects.create(name='Coke', colour='Black')
<Drink: Coke>

>>> HotDrink.objects.create(name='Coffee', colour='Black')
<HotDrink: Coffee, really gets you going!>

>>> AlcoholicDrink.objects.create(name='Beer', colour='Golden', strength=6.7)
<AlcoholicDrink: Beer>
```

And then I query the various sets:

```
>>> Drink.objects.all()
[<Drink: Water>, <Drink: Coke>, <HotDrink: Coffee, really gets you going!>, <AlcoholicDrink: Beer>]

... correctly returns all the drinks with their types resolved

>>> AlcoholicDrink.objects.all()
[<AlcoholicDrink: Beer>]

... correctly returns the sub-types from AlcholicDrinks
```

However, if I query on the 'HotDrinks' type, I get all of the 'Drinks' instead of **just** the 'HotDrinks':
```
>>> HotDrink.objects.all()
[<Drink: Water>, <Drink: Coke>, <HotDrink: Coffee, really gets you going!>, <AlcoholicDrink: Beer>]
```

I can get around this by using the following:
```
>>> HotDrink.objects.instance_of(HotDrink)
[<HotDrink: Coffee, really gets you going!>]
```

Is this an issue (it seems like an issue to me)? Or is it by design?

andrewdodd:
I'm happy to look into fixing it if it is, just keen to not waste time if it isn't an issue.

vdboor:
[...]

PR #210, opened 2016-05-04
Issue #200 - Exclude super-class from proxy model query sets

andrewdodd:
These changes prevent the query sets on 'proxy' models from including
the items from the 'super-class'.

I'm not really sure if it is as simple as these change suggest, but...


vdboor:
@andrewdodd that is a brilliant fix! It could really be as simple as this.

PR #211, opened 2016-05-05
Update admin.rst

nikolaz111:
I found this to be useful when we want a base model  object to be used in the admin


vdboor:
It's surely useful to have. However, the method for registering models will change in the next version (with this style being a backwards compatible way). You can register the child admins in the regular way now, and they have a `show_in_index` flag that can be set.

PR #215, opened 2016-05-20
Admin refactoring (based on #58)

skirsdeda:
I took #58 and reimplemented it on current HEAD without removing any of the old code. This means that it is fully backwards compatible.

I am not sure what to do with DeprecationWarning which I added because python hides this by default and I couldn't find any way to make it visible when running the usual ```manage.py runserver```.

Also let me know if I should add something to Changelog. Although text from #58 could be reused I guess.

### DIFF
--- a/polymorphic/admin.py
+++ b/polymorphic/admin.py
@@ -2,14 +2,17 @@
 ModelAdmin code to display polymorphic models.
 """
 import sys
+import warnings
+
+import django
 from django import forms
 from django.conf.urls import url
 from django.contrib import admin
-from django.contrib.admin.helpers import AdminForm, AdminErrorList
+from django.contrib.admin.helpers import AdminErrorList, AdminForm
 from django.contrib.admin.widgets import AdminRadioSelect
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import PermissionDenied
-from django.core.urlresolvers import RegexURLResolver
+from django.core.urlresolvers import RegexURLResolver, resolve
 from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import render_to_response
 from django.template.context import RequestContext
@@ -18,8 +21,6 @@ from django.utils.encoding import force_text
 from django.utils.http import urlencode
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
-import django
-
 
 try:
     # Django 1.6 implements this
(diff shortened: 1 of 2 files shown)


skirsdeda:
Also added ```show_in_index``` attribute to ```PolymorphicChildModelAdmin``` just in case someone wants to see child model separately in the admin index.

polymorphic/admin.py
```diff
@@ -136,6 +135,23 @@ def _lazy_setup(self):
         # By not having this in __init__() there is less stress on import dependencies as well,
         # considering an advanced use cases where a plugin system scans for the child models.
         child_models = self.get_child_models()
+        # Check if child_models are provided via new syntax (iterable of models) or compatibility syntax
+        # (iterable of (Model, Admin). When no child_models are provided, assume new syntax.
+        self._compat_mode = len(child_models) and isinstance(child_models[0], (list, tuple))
```
vdboor: there can be a difference between `self.child_models` and `self.get_child_models()` - similar to the way Django views have a `get_form_class()` and `form_class` attribute. It's recommended to always use `get_child_models()` as some packages (e.g. django-fluent-pages) can provide completely dynamic input here.

vdboor:
Thank you for making a very clear pull request to achieve what #58 did while preserving backwards compatibility. I've made a small comment in the diff, you may want to review that.

polymorphic/admin.py
skirsdeda: Of course. It's just a matter of naming in this case. I'll change it.

skirsdeda:
@vdboor Any news on this? Let me know if anything needs to be done to merge this.

vdboor:
@skirsdeda I've updated the example project too to handle the new-style registration.
I've fixed one bug (c1ee5aad0b66e1dcd1554081353786dbacb16515) and one other remains: when an object is added, the user is redirected to the child-admin, and not the parent admin. Could you look into this?

skirsdeda:
@vdboor Sure, I'll look into that.

skirsdeda:
@vdboor It seems that this could be solved by adding ```response_post_save_add``` and ```response_post_save_change``` methods on ```PolymorphicChildAdmin``` which would find parent admin instance and call those methods on the parent. This would be compatible down to Django **1.5**. If that's fine, I could create a pull request for such fix.

vdboor:
@skirsdeda that is a great idea, I'd love to receive a PR for this.

skirsdeda:
@vdboor While trying to implement that, I hit another problem with new admin implementation. Previously, it was possible to use parent polymorphic model in ```child_models``` so that one gets choice of parent model in add page, now it's impossible because admin site doesn't allow registering two admins for one model. So I guess some flag on ```PolymorphicParentAdmin``` could be added to signify that parent model should be available as child model (to choose in add page and in filter by polymorphic type). I'll incorporate this in the same PR.

Issue #216, opened 2016-05-28
Unable to save polymorphic models to non-default databases


rp-austin:
Saving subclasses of `PolymorphicModel` to databases other than `default` fails due to an error in the `PolymorphicModel.pre_save_polymorphic` method.

**Steps to reproduce:**

1. Configure the `example` Django project to use two databases. For example, change the `DATABASES` entry in `example/settings.py` to:
    ```python
    DATABASES = {
        'default': {
            'ENGINE': 'django.db.backends.sqlite3',
            'NAME': os.path.join(PROJECT_ROOT, 'example.db'),
        },
        'secondary': {
            'ENGINE': 'django.db.backends.sqlite3',
            'NAME': os.path.join(PROJECT_ROOT, 'example2.db'),
        }
    }
    ```

2. Leave the `default` database empty. Migrate the project using the `secondary` database:

    ```bash
    python manage.py migrate --database secondary
    ```

3. Attempt to save an instance of a `PolymorphicModel` subclass using the `secondary` database:

    ```python
    from pexp.models import Model2A
    Model2A(field1='test').save(using='secondary')
    # or...
    Model2A.objects.db_manager('secondary').create(field1='test')
    ```

**Expected result:**

The instance is saved into the `secondary` database.

**Observed result:**

Saving fails with the following traceback:

```python
Traceback (most recent call last):
  File "/Users/austin/.virtualenvs/polymorphic/lib/python3.5/site-packages/django/core/management/commands/shell.py", line 69, in handle
    self.run_shell(shell=options['interface'])
  File "/Users/austin/.virtualenvs/polymorphic/lib/python3.5/site-packages/django/core/management/commands/shell.py", line 61, in run_shell
    raise ImportError
ImportError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/austin/.virtualenvs/polymorphic/lib/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/austin/.virtualenvs/polymorphic/lib/python3.5/site-packages/django/db/backends/sqlite3/base.py", line 323, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.OperationalError: no such table: django_content_type

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #217, opened 2016-05-28
#216 Allow ContentType queries to be performed on non-default databases.

rp-austin:


vdboor:
Thanks for the wonderful work, this seems like a ready to merge feature!

I wonder, instead or passing `using=DEFAULT_DB_ALIAS` to all model methods, would it work to read `sefl._state.db` or is the `ContentType` located in a different DB then the models that derive from it?

rp-austin:
Thanks for the feedback! I took your suggestion and switched to using `self._state.db` in the model methods. It keeps the interface cleaner and I think is more consistent with Django's behavior, so it seems like a better way to handle this.

(I didn't actually know about `ModelState` before your comment, which is why I wasn't using it before :) )

The one exception is the `save` method, and by extension the `pre_save_polymorphic` method. `save` checks the `using` kwarg, then `self._state.db`, then falls back on `DEFAULT_DB_ALIAS`; `pre_save_polymorphic` simply uses the kwarg value. Since `self._state.db` won't be populated for new instances and callers may also override the database when calling `save`, I thought it was better to use the argument. However, I can certainly change this behavior if you'd prefer.

I also fixed a couple spots in the `query_translate` functions where I wasn't passing the `using` value correctly before. `instance_of` filters should work correctly now.

Finally, I've also added a couple unit tests for the new functionality, but not many. If I need to add more, please let me know.

Thanks!

vdboor:
Wonderful, thanks for updating this swiftly!

Issue #213, opened 2016-05-11
Don't Mutate Q Objects in Polymorphic Query _filter_or_exclude


alexander-alvarez:
I'm using a global (what I thought was constant) Q object, and it turns out that if you reuse it bad things may happen.

**The culprit**
https://github.com/chrisglass/django_polymorphic/blob/master/polymorphic/query.py#L115

**Potential fix**
Follow the paradigm of  [Django's QuerySet._filter_or_exclude](https://github.com/django/django/blob/7ec330eeb96d0874949eacb8ed1bbb97e11807e1/django/db/models/query.py#L802) where the Q objects are copied. In our case we'd copy them before transforming them, that way the original object is not modified.

**TestCase**
```py

def test_query_filter_exclude_is_immutable(self):
    # given 
    q_to_reuse = Q(Model2B___field2='something')
    untouched_q_object = Q(Model2B___field2='something')
    # when
    Model2A.objects.filter(q_to_reuse).all()
    # then

    # will fail now because q_to_reuse.children[0] is now ('field2', 'something')
    # and will fail if used again since field2 doesn't exist in Model2A
    self.assertListEquals(q_to_reuse.children, untouched_q_object.children)

    #given
    q_to_reuse = Q(Model2B___field2='something')
    untouched_q_object = Q(Model2B___field2='something')
    # when
    Model2B.objects.filter(q_to_reuse).all()
    # then
    self.assertListEquals(q_to_reuse.children, untouched_q_object.children)

```

vdboor:
Thanks for the test case! I see what polymorphic is causing here. I'm not available right now for making fixes, but if anyone can provide a PR I can review and merge that!

PR #218, opened 2016-06-02
Issue #213: Don't modify Q passed in as arguments

alexander-alvarez:
closes https://github.com/chrisglass/django_polymorphic/issues/213

It's a fairly simple PR. I haven't tried to investigate how to clone Q objects in Django 1.4/1.5 yet. Any ideas?


vdboor:
@alexander-alvarez thanks for this change! I wonder whether we should still support Django 1.4/1.5 on this. We could add an `if django.VERSION < (1, 6)` to the translate function, making this feature 1.6+ only.

alexander-alvarez:
@vdboor Updated tests with the flag

vdboor:
Thanks a lot!

PR #221, opened 2016-06-13
#220 Pass hints to returned queryset in PolymorphicManager.get_queryset.

### LINKED ISSUE #220: PolymorphicManager.get_queryset() does not pass hints to returned queryset
`PolymorphicManager.get_queryset()` is defined as follows:

```python
def get_queryset(self):
    qs = self.queryset_class(self.model, using=self._db)
    if self.model._meta.proxy:
        qs = qs.instance_of(self.model)
    return qs
```

Any hints defined on the manager are not passed to the new queryset. This can be a problem when it's not possible to specify a database on the manager itself or when relying on database routers.

`PolymorphicManager.get_queryset()` should be updated to include hints:

```python
def get_queryset(self):
    qs = self.queryset_class(self.model, using=self._db, hints=self._hints)
    if self.model._meta.proxy:
        qs = qs.instance_of(self.model)
    return qs
```

Additionally, `PolymorphicQuerySet` should be updated to use the `self.db` property instead of `self._db` so database routers are used when `self._db` isn't set.

--

An example of when this is a problem is when accessing objects via foreign key properties. E.g.:

```python
class Parent(Model):
    pass

class Child(Model):
    parent = ForeignKey(Parent, related_name='children')

Child.objects.create(parent=Parent())
Child.objects.first().parent
```

This is because Django's `ForwardManyToOneDescriptor.__get__()` uses the following `ForwardManyToOneDescriptor.get_queryset()` method, passing an object instance as a hint:

```python
def get_queryset(self, **hints):
    manager = self.field.remote_field.model._base_manager
    # If the related manager indicates that it should be used for
    # related fields, respect that.
    if not getattr(manager, 'use_for_related_fields', False):
        manager = self.field.remote_field.model._base_manager
    return manager.db_manager(hints=hints).all()
```

The `PolymorphicQuerySet` then has no knowledge of any other databases, so any queries it performs invariably use the `default` database. This can result in incorrect objects being returned if the databases have the same schema, or database errors otherwise.

rp-austin:
- Pass hints to returned queryset in `PolymorphicManager.get_queryset()`.
- Use `self.db` instead of `self._db` in `PolymorphicQuerySet`.
- Add unit tests for related object queries using non-default databases.

Fixes #220.

### DIFF
--- a/polymorphic/managers.py
+++ b/polymorphic/managers.py
@@ -32,7 +32,10 @@ class PolymorphicManager(models.Manager):
         super(PolymorphicManager, self).__init__(*args, **kwrags)
 
     def get_queryset(self):
-        qs = self.queryset_class(self.model, using=self._db, hints=self._hints)
+        if django.VERSION >= (1, 7):
+            qs = self.queryset_class(self.model, using=self._db, hints=self._hints)
+        else:
+            qs = self.queryset_class(self.model, using=self._db)
         if self.model._meta.proxy:
             qs = qs.instance_of(self.model)
         return qs
--- a/polymorphic/managers.py
+++ b/polymorphic/managers.py
@@ -32,7 +32,7 @@ class PolymorphicManager(models.Manager):
         super(PolymorphicManager, self).__init__(*args, **kwrags)
 
     def get_queryset(self):
-        qs = self.queryset_class(self.model, using=self._db)
+        qs = self.queryset_class(self.model, using=self._db, hints=self._hints)
         if self.model._meta.proxy:
             qs = qs.instance_of(self.model)
         return qs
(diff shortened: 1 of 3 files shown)


rp-austin:
I didn't realize hints weren't available prior to Django 1.7. I'll add a version check and push the changes shortly.

vdboor:
Thanks! These are wonderful changes! I wonder whether you could make `test_save_to_non_default_database()` Django 1.4 compatible, or add a `@skipIf(..)` condition as 1.4 is getting very old :P

rp-austin:
Thanks! I ended up simply skipping the failing test for 1.4, though I'm happy to investigate it further if it's an issue.

vdboor:
Great, it's all merged in now.

PR #222, opened 2016-06-14
Fix redirects after child model add/change in PolymorphicChildModelAdmin

skirsdeda:
Since it's possible to have parent model also as a child model in ```child_models``` list, in compatibility mode ```_get_parent_admin``` will do a roundtrip in such case (first it will find parent model's PolymorphicChildAdmin, and on the second call will return super of that PolymorphicChildAdmin subclass). So when in compatibility mode and parent model is among child models, ```_get_parent_admin``` will return super of child admin (not super of parent admin, which happens in all other cases), but it works fine because response_post_save methods redirect to same urls :)

### DIFF
--- a/polymorphic/admin/childadmin.py
+++ b/polymorphic/admin/childadmin.py
@@ -106,6 +106,20 @@ class PolymorphicChildModelAdmin(admin.ModelAdmin):
             "admin/object_history.html"
         ]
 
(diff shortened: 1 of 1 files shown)


vdboor:
Wonderful, thanks for fixing this!

PR #223, opened 2016-06-15
Convert readthedocs links for their .org -> .io migration for hosted projects

adamchainz:
As per [their blog post of the 27th April](https://blog.readthedocs.com/securing-subdomains/) ‘Securing subdomains’:

> Starting today, Read the Docs will start hosting projects from subdomains on the domain readthedocs.io, instead of on readthedocs.org. This change addresses some security concerns around site cookies while hosting user generated data on the same domain as our dashboard.

Test Plan: Manually visited all the links I’ve modified.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -52,7 +52,7 @@ Django to perform an ``INNER JOIN`` to fetch the model fields from the database.
 While taking this in mind, there are valid reasons for using subclassed models.
 That's what this library is designed for!
 
-For more information, see the `documentation at Read the Docs <https://django-polymorphic.readthedocs.org/>`_.
+For more information, see the `documentation at Read the Docs <https://django-polymorphic.readthedocs.io/>`_.
 
 License
 =======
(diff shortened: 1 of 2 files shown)


adamchainz:
The link on the Github repo description also needs updating.

vdboor:
Thanks!

vdboor:
@chrisglass: could you update the repo link, or give me access to this?

chrisglass:
Hmm I thought you'd have full access to everything.

I just manually updated the repo link for now, but it's surprising that you don't have full access. Perhaps we should make a github team instead of keeping this in my namespace (I've not looked at or otherwise worked with the code forever).

Thoughts?

adamchainz:
Github permissions  ¯\_(ツ)_/¯

vdboor:
@chrisglass I'm in favor of changing that! It's also more encouraging for people to join, because there is more a sense of a team instead of a single person doing everything,

chrisglass:
Let's try to discuss on irc today (in an hour or two)? #django-polymorphic
on freenode? I think it's faster/better this way :)

vdboor:
@chrisglass : perfect!

Issue #225, opened 2016-06-29
Compatibility with Django 1.10: Attribute error


PetrDlouhy:
When running django-polymorphic on upcoming Django 1.10, following error is raised:
```
Traceback (most recent call last):
  File "runtests.py", line 83, in <module>
    runtests()
  File "runtests.py", line 80, in runtests
    execute_from_command_line(argv)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/core/management/__init__.py", line 341, in execute
    django.setup()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/__init__.py", line 27, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/apps/registry.py", line 108, in populate
    app_config.import_models(all_models)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/apps/config.py", line 206, in import_models
    self.models_module = import_module(models_module_name)
  File "/opt/python/2.7.9/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/travis/build/PetrDlouhy/django-polymorphic/polymorphic/models.py", line 31, in <module>
    class PolymorphicModel(six.with_metaclass(PolymorphicModelBase, models.Model)):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/utils/six.py", line 808, in __new__
    return meta(name, bases, d)
  File "/home/travis/build/PetrDlouhy/django-polymorphic/polymorphic/base.py", line 90, in __new__
    new_class._default_manager = user_manager._copy_to_model(new_class)
AttributeError: 'PolymorphicManager' object has no attribute '_copy_to_model'
```

JasonYao:
Commenting here as well to say this issue is still persisting as of Aug 2nd.

uWSGI log:
> *** Operational MODE: preforking ***
Traceback (most recent call last):
  File "./Test/wsgi.py", line 16, in <module>
    application = get_wsgi_application()
  File "/server/.env/lib/python3.5/site-packages/django/core/wsgi.py", line 13, in get_wsgi_application
    django.setup(set_prefix=False)
[...]

PR #226, opened 2016-06-29
Test in development Django version (with allow failures)

PetrDlouhy:
This patch runs enables Travis tests in Django development (with allow failures) to show the errors of current codebase with Django 1.10 (i.e. #225).
It also contains correction for new templates handling in tests in Django 1.10


vdboor:
Thanks!

PR #228, opened 2016-07-07
Update admin.rst

mekhami:
child_models should be an iterable of (Model, ModelAdmin) pairs.


vdboor:
This will change in the 1.0 release. I've updated readthedocs to point to the stable (v0.9.2) docs.
See: http://django-polymorphic.readthedocs.io/en/stable/admin.html

PR #230, opened 2016-07-29
Fix package name

michael-borisov:


vdboor:
We really like to move forward to the dash notation instead. Our GitHub repo name also changed. Where does using the dash affect you?

PR #231, opened 2016-07-29
Docs for PolymorphicParentModelAdmin are wrong

rfleschenberg:
The admin docs contradict the code, and also the docstring of `PolymorphicParentModelAdmin`. This should fix it.


vdboor:
This change is intentional, it will be part of the next release. However, did you still see it on RTD? We've fixed that in #228

rfleschenberg:
Yes, I saw it on RTD. Maybe I was looking at the "latest" docs for some reason.

Issue #227, opened 2016-07-07
TypeError: 'PolymorphicModelBase' object is not iterable


mekhami:
http://dpaste.com/2Z3RH41

`TypeError: 'PolymorphicModelBase' object is not iterable`

I am getting this error when trying to set up the admin for my polymorphic models. Here's a link to the admin for those models: https://gist.github.com/mekhami/07288765ad1b6d72ed4bda10022a0581

And here's a link to the models themselves: https://gist.github.com/mekhami/c27333c7c217e1da07f618f8a29ed517

Any insight? I think I've followed the documentation to a letter..

mekhami:
Ah, so the documentation is a bit outdated. child_models should be an iterable of (Model, ModelAdmin) tuples.

So there's a documentation fix to be made here anyway.

glifchits:
Bump: the docs should get updated!

PR #232, opened 2016-08-12
Update docs: Admin child_models are (Model, ModelAdmin) pairs

glifchits:
If users follow the documentation, they'll run into issue #227 (like I did).


vdboor:
When you open the 'stable' documentation (the default), it does show the pairs:
http://django-polymorphic.readthedocs.io/en/stable/admin.html

The 'latest' documentation now highlights that this will change between 0.9 and 1.0:
http://django-polymorphic.readthedocs.io/en/latest/admin.html

This duplicates #231 #228 - I wonder how you landed on the 'latest' documentation. Was that a direct link?

glifchits:
I found the docs via google, didn't think to check the version. Sorry about that!

PR #238, opened 2016-09-03
CSS tweak for Grappelli

Quantra:
Prevent div.polymorphic-add-choice overlapping previous inline formset when using Grappelli.

Also tested without Grappelli.

screen shot 2016-09-03 at 15 17 53


vdboor:
Thanks!

PR #241, opened 2016-09-20
Improve django-guardian integration docs.

benkonrath:
There was a mistake with the `GUARDIAN_GET_CONTENT_TYPE` setting in the django-guardian integration docs. I've also re-worked the explanation to make it more clear.


vdboor:
Thanks!

PR #242, opened 2016-09-21
Wrap call to key() in list() so it can be indexed

chrisbrantley:
In Python 3.4 trying to generate a polymorphic_inlineformset results in a TypeError with the following message: "'KeysView' object does not support indexing". This solves that problem by ensuring that `child_models` is a list, and thus can be referenced by index.


vdboor:
Thanks! :D

PR #246, opened 2016-10-12
Fix get parent admin when intermediate not root ctype model is registered

jmurty:
This change fixes an issue where django-polymorphic raises a
`ParentAdminNotRegistered` exception when you register an admin for a
child model ancestor, but not for the root ancestor as pointed to by
the `polymorphic_ctype` field. This error occurs only when you *Save*
the child model detail form, not when you *Save and continue* on that
same form.

This situation occurs for us when using django-fluent-pages version
1.0.1 which has an intermediate `Page` model registered with a parent
admin to show the pages listing. The existing `_get_parent_admin` method
expects an admin to be registered for the root `UrlNode` model pointed
to by the `polymorphic_ctype` field.

This fix uses a potentially naive and slow brute-force approach where
it walks up the class hierarchy and checks  whether a parent admin is
registered for each ancestor model, unless/until it finds one.

See also https://github.com/ic-labs/django-icekit/issues/31/

### DIFF
--- a/polymorphic/admin/childadmin.py
+++ b/polymorphic/admin/childadmin.py
@@ -1,12 +1,15 @@
 """
 The child admin displays the change/delete view of the subclass model.
 """
+import inspect
+
 from django.contrib import admin
 from django.core.urlresolvers import resolve
 from django.utils import six
 from django.utils.translation import ugettext_lazy as _
 
 from .helpers import PolymorphicInlineSupportMixin
+from ..admin import PolymorphicParentModelAdmin
 
 
 class ParentAdminNotRegistered(RuntimeError):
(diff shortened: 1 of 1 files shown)


jmurty:
Let me know if the fix is on the right track and potentially mergeable, and if so I will improve the test code coverage. I will hold off on the unit tests for now in case I'm doing completely the wrong thing.

vdboor:
Thanks for this fix! Astonishing I didn't see this error before - it really confirms I always use the "Save and Continue" button in the admin!

jmurty:
Thanks for the merge. And I know what you mean, I needed to be told a few times myself before I understood that the *Save* action triggered an error, because I never use it either.

vdboor:
I've brushed up the code a bit in 58f89efa5d2928e0a5568d6100abe5495e2d40bf, I hoe you don't mind. It also helped me understanding what's going on, and be more explicit that it's an intermediate/proxy model the code is looking for.

jmurty:
Not at all, I wasn't confident in the comments' terminology and your changes clarify things. I'll work on improved unit tests next week

Issue #68, opened 2014-02-21
Prefetch_related has strange behavior


EliotBerriot:
Hi,

I use django-polymorphic for a personal, non-critical project, in which items (Storable) can be stored in workspaces (Workspace). 

This project relies heavily on concrete inheritance, and my models all inherit from a base model (using Polymorphic).

Everything works fine but, as always with concrete inheritance, database load is quite high, especially when I want to display a list of Storable, along with their workspaces.
In order to reduce the number of queries, I'm trying to use `prefetch_related('workspaces')` on Storable queryset. However, while it reduces indeed the number of queries, there is a bug when come the time to display items.

I've set up a test project with the following code.
# Models

```
from django.db import models
from polymorphic import PolymorphicModel


class Base(PolymorphicModel):
    name = models.CharField('name', max_length=255, default=None, blank=True, null=True)


class Storable(Base):
    pass


class Workspace(Base):
    items = models.ManyToManyField(Storable, related_name="workspaces", null=True, blank=True, default = None)
```
# View

```
from django.views.generic.list import ListView
from test_prefetch_related.models import Storable


class TestPrefetchRelated(ListView):
    model = Storable
    template_name = "test_prefetch_related/test_prefetch_related.html"
    paginate_by = 10

    def get_queryset(self):
        return self.model.objects.all().prefetch_related('workspaces')
```
# Template

```
<!DOCTYPE html>
<html>
<head>
    <title></title>
</head>
<body>
    {% for item in object_list %}
        <h1>{{ item.name }}</h1>
        <ul>
            {% for w in item.workspaces.all %}
                <li>{{ w.id }} - {{ w.name }}</li>
            {% empty %}
                None
            {% endfor %}
        </ul>
    {% endfor %}
</body>
</html>
```
# Database population

```
from test_prefetch_related.models import Storable, Workspace

# here is the code I used to populate the database with some data

# create workspaces
w1 = Workspace(name="w1")
w1.save()

w2 = Workspace(name="w2")
w2.save()

w3 = Workspace(name="w3")
w3.save()

# create storables
s1 = Storable(name="s1")
s1.save()
s1.workspaces.add(w1)
s1.save()

s2 = Storable(name="s2")
s2.save()
s2.workspaces.add(w2)
s2.save()

s3 = Storable(name="s3")
s3.save()
s3.workspaces.add(w1, w2, w3)  # this storable get all workpaces
s3.save()
```
# Html output

```
s1

    1 - w1
[...]

PR #250, opened 2016-10-19
Added sample test code to reproduce issue #68

EliotBerriot:
This is just an attempt to deal with #68.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ build/
 dist/
 docs/_build/
 htmlcov/
+venv/
--- a/polymorphic/tests.py
+++ b/polymorphic/tests.py
@@ -428,6 +428,14 @@ class DateModel(PolymorphicModel):
     date = models.DateTimeField()
 
 
+class PrefetchRelatedA(PolymorphicModel):
+    pass
+
+
+class PrefetchRelatedB(models.Model):
+    links = models.ManyToManyField(PrefetchRelatedA)
+
+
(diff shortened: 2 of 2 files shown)


EliotBerriot:
I took some time yesterday to track down the issue, and here is what I got:

1. Apparently, `prefetch_related` on the queryset calls `prefetch_related_objects` (https://github.com/django/django/blob/1.10.1/django/db/models/query.py#L1387)
2. This function calls `prefetch_one_level` (https://github.com/django/django/blob/1.10.1/django/db/models/query.py#L1473)
3. At some point, this function retrieve a queryset from the prefetcher `get_prefetch_queryset` instance (https://github.com/django/django/blob/1.10.1/django/db/models/query.py#L1580). If I'm right, this queryset of related objects [is then looped](https://github.com/django/django/blob/1.10.1/django/db/models/query.py#L1602) over and results are binded to original requested instances
4. The mess seems to happen exactly [here](https://github.com/django/django/blob/1.10.1/django/db/models/fields/related_descriptors.py#L853): to retrieve the related objects, an `extra` query is used by the `get_prefetch_queryset` method, and it is in this method I found the wrong PK's of the related objects to be returned.

I'll take more time to clearly identify if the line linked in 4 is where the error happens, because this ORM code is full of lambdas, descriptors and other deep-vodoo-black-magic stuff that are not easy to grasp. However, this `extra` query seems like a good culprit to me.

PR #254, opened 2016-12-09
Add a failing test to demonstrate a bug

mjnaderi:
We found a bug in our project, and after investigating, we found that where it comes from. So I created a failing test based on our issue.

In this test, removing `prefetch_related` from this line:

```python
my_teams = Team.objects.filter(user_profiles=user_c).prefetch_related('user_profiles').distinct()
```

or removing this "for" loop:

```python
for _ in my_teams:
    pass
```

resolves the issue, and test passes. But with both of them, test fails.

### DIFF
--- a/polymorphic/tests.py
+++ b/polymorphic/tests.py
@@ -167,6 +167,22 @@ class ModelUnderRelChild(PolymorphicModel):
     _private2 = models.CharField(max_length=10)
 
 
+class Participant(PolymorphicModel):
+    pass
+
+
+class UserProfile(Participant):
+    name = models.CharField(max_length=100)
+
+    def __str__(self):
+        return self.name
+
+
+class Team(models.Model):
+    team_name = models.CharField(max_length=100)
+    user_profiles = models.ManyToManyField(UserProfile, related_name='user_teams')
+
+
 class MyManagerQuerySet(PolymorphicQuerySet):
 
     def my_queryset_foo(self):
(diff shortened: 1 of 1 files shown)


mjnaderi:
Another note:
The behavior is different with Sqlite and PostgreSql.

With Sqlite, the test fails with:

```
  File "/home/naderi/Repos/django-polymorphic/polymorphic/tests.py", line 1379, in test_unknown_issue
    self.assertEqual(len(my_teams[0].user_profiles.all()), 3)
AssertionError: 4 != 3
```

With PostgreSql, the test fails with:

```
  File "/home/naderi/Repos/django-polymorphic/polymorphic/tests.py", line 1379, in test_unknown_issue
    self.assertEqual(len(my_teams[0].user_profiles.all()), 3)
AssertionError: 2 != 3
```

mjnaderi:
Now I found issue #68 and pull request #250. Maybe this issue is related to them.

polymorphic/tests.py
```diff
@@ -1325,6 +1341,44 @@ def func():
...
+        team2 = Team.objects.create(team_name='team2')
+        team2.user_profiles.add(user_c)
+        team2.save()
+
+        # without prefetch_related, the test passes
+        my_teams = Team.objects.filter(user_profiles=user_c).prefetch_related('user_profiles').distinct()
+
+        print(my_teams[0].user_profiles.all())
+        print(my_teams[1].user_profiles.all())
+        self.assertEqual(len(my_teams[0].user_profiles.all()), 3)
```
WhyNotHugo: This doesn't look right; you're not ordering you QuerySet, so the first result returned by the DB might be `team2`. It'll probably fail, occasionally.

bckohan:
This PR has been stuck in this state for the last day, so I have pulled the commits and am superceding it with #671 
Screenshot 2025-12-03 at 11 00 48 AM

PR #259, opened 2017-01-30
Fix query resolving when child's model have changed related_name

petyunchik:
When django-polymorphic constructs field paths for querying it doesn't count that child's model may have redefined link to parent class with changed related_name.

This pull request improves path resolving using django's native capabilities.


vdboor:
Thanks, this looks like a great starting point! However, notice that the tests break because of it. Your solution isn't compatible with various Django versions. Could you look into this>

vdboor:
I've merged this fix!

Issue #263, opened 2017-02-07
TypeError: Can't convert 'bytes' object to str implicitly


DDevine:
When creating a new object via Django Admin which triggers a pop-up interface I ran into this Python3 incompatible implicit type conversion. 
```
Request Method: 	GET
Request URL: 	http://127.0.0.1/admin/example/scene/add/?_to_field=id&_popup=1
Django Version: 	1.10.5
Exception Type: 	TypeError
Exception Value: 	

Can't convert 'bytes' object to str implicitly

Exception Location: 	/opt/example-env/lib/python3.5/site-packages/polymorphic/admin/parentadmin.py in add_type_view, line 362
Python Executable: 	/opt/example-env/bin/python
Python Version: 	3.5.2
````
The problematic line is:
```
extra_qs = ''
if request.META['QUERY_STRING']:
    extra_qs = '&' + request.META['QUERY_STRING']
```

The value of META[QUERY_STRING] is:
```b'_to_field=id&_popup=1'```
Which as you can see is a byte object, and it is being concatinated with the '+' operator which is trigging an invalid implicit conversion in Python 3. The fix I used is to use string substitution...
```
extra_qs = "&%s" % request.META['QUERY_STRING']
```

PR #264, opened 2017-02-07
Changed concatination method to successfully perform 'bytes' to 'str'…

DDevine:
This is a fix for #263


vdboor:
Thanks for highlighting this issue. I looked into this, and made a different fix in 1694699debd9a7b8be6fa03d8b54edcb361bc906

My reasoning is that Python 3 makes the difference between `bytes` and `str` for a reason. The input is not encoded yet to unicode, so merging it with a unicode string could cause other errors for unicode input. Django's `QueryDict` class takes a similar approach.

DDevine:
Not tested yet, but thanks!

PR #268, opened 2017-03-01
Replace force_unicode with force_text for Python3 compatibility.

fladi:
`force_unicode` has been [deprecated](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.encoding.force_unicode) and replaced with `force_text`[0].
Furthermore, `force_unicode` is not available in Python3, causing the
documentation to be unbuildable with Python3 sphinx.


vdboor:
Thanks!

PR #269, opened 2017-03-01
Install django.contrib.contenttypes for docstrings.

fladi:
This is merely cosmetic to avoid Django warning because of
`django.contrib.contenttypes` being referenced but not an installed app.


vdboor:
Thanks!

PR #276, opened 2017-04-12
WIP: Get tests running on django 1.11

meshy:
I'm a little concerned that this loses some of the efficiencies (in particular, chunking) from previous versions. That's something that can surely be improved.

This addresses #267.

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -25,6 +25,25 @@ except ImportError:
 Polymorphic_QuerySet_objects_per_request = CHUNK_SIZE
 
 
+if django.VERSION >= (1, 9):
+    # We ignore this on django < 1.9, as ModelIterable didn't yet exist.
+    from django.db.models.query import ModelIterable
+
+
+    class PolymorphicModelIterable(ModelIterable):
+        def __iter__(self):
+            base_iter = super(PolymorphicModelIterable, self).__iter__()
+
+            if self.queryset.polymorphic_disabled:
+                for o in base_iter:
+                    yield o
+                return
+
+            real_instances = self.queryset._get_real_instances(base_iter)
+            for obj in real_instances:
+                yield obj
+
+
 def transmogrify(cls, obj):
     """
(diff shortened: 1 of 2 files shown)


polymorphic/query.py
```diff
@@ -25,6 +25,25 @@
...
+            base_iter = super(PolymorphicModelIterable, self).__iter__()
+
+            if self.queryset.polymorphic_disabled:
+                for o in base_iter:
+                    yield o
+                return
+
+            real_instances = self.queryset._get_real_instances(base_iter)
+            for obj in real_instances:
+                yield obj
```
meshy: Should we be chunking with `Polymorphic_QuerySet_objects_per_request`?

Django 1.11 introduced [server-side cursors](https://docs.djangoproject.com/en/1.11/releases/1.11/#database-backends) for postgres... I have no idea how this would play with that.

meshy:
I'd really like to try this on postgres and mysql as well as sqlite... I'll open another PR to add those envs to tox...

polymorphic/query.py
vdboor: > Should we be chunking with Polymorphic_QuerySet_objects_per_request?
Yes, you'll need that. Suppose you read 10+ objects from an iterator, but the actual table is 1000+ objects. Without chunking you'd retrieve all those objects, and then the iterator walks over it.

polymorphic/query.py
```diff
@@ -25,6 +25,25 @@
...
+
```
vdboor: I don't think so, omitting it would be an implicit `return None` effectively

vdboor:
Thanks for the work on this PR!

I've merged this with a combination of #279 that handled the chunking in both cases too.

meshy:
Excellent, thank you!

I was a little nervous about the chunking because of django's own changes in that regard, but #279 seems like a good way forward to me.

If I get the time, I hope to be able to get the tests running against mysql and postgres soon. I'll open another PR for that, of course. Thanks again for the merge `:)`

PR #277, opened 2017-04-14
Use tox for travis builds

meshy:
Before adding postgres and mysql builds to travis, I thought it would be a good idea to get this out of the way first, as it'll make things easier (and more DRY) for adding postgres and mysql tests.


.travis.yml
```diff
@@ -10,97 +10,96 @@ python:
...
+    - python: "3.2"
+      env: TOX_ENV="django19"
+    - python: "3.2"
+      env: TOX_ENV="django110"
+    - python: "3.2"
+      env: TOX_ENV="django111"
+    - python: "3.2"
+      env: TOX_ENV="djangodev"
+
+    - python: "2.6"
```
hobarrera: Since `pip` is about to drop support for 2.6, it's probably a good time to get rid of it.

.travis.yml
```diff
@@ -10,97 +10,96 @@ python:
...
 before_install:
-- pip install codecov
-
-install:
-  - pip install -q $DJANGO coverage==3.6
+- pip install codecov tox
 
 script:
-  - coverage run --source=polymorphic runtests.py
+  - tox -e py${TRAVIS_PYTHON_VERSION/./}-$TOX_ENV
```
hobarrera: You don't need to pass `$TOX_ENV`, and since you're changing the default system python on the travis environment, you don't need to specify that either.

Simply `- tox` is enough here.

tox.ini
```diff
@@ -21,9 +20,10 @@ deps =
     django19: Django >= 1.9, < 1.10
     django110: Django >= 1.10, < 1.11
     django111: Django >= 1.11, < 1.12
-    django-dev: https://github.com/django/django/tarball/master
+    djangodev: https://github.com/django/django/tarball/master
 commands=
-    python runtests.py
+    coverage run --source polymorphic runtests.py
+    coverage xml -i
```
hobarrera: Why generate the XML here? AFAIK, codecov doesn't need that explicitly (or am I missing something?).

.travis.yml
meshy: You're right. I also think it's worth dropping support for old versions of django. That's one for another issue though, as that will have an impact on the main code, and I'd like to limit the scope of this issue only to moving travis to tox.

tox.ini
meshy: I did not expect to have to do this for codecov, but I noticed that without this, codecov was printing the following message:
```
==> Detecting CI provider
    Travis Detected
==> Preparing upload
==> Processing gcov (disable by -X gcov)
    Executing gcov (find /home/travis/build/django-polymorphic/django-polymorphic -type f -name '*.gcno'  -exec gcov -pb  {} +)
==> Collecting reports
    Generating coverage xml reports for Python
    Error running `coverage xml -i`: None
    - Ignored: [Errno 2] No such file or directory: '/home/travis/build/django-polymorphic/django-polymorphic/coverage.xml'
Error: No coverage report found
```
After adding this line, the message changed...

```
==> Detecting CI provider
    Travis Detected
==> Preparing upload
==> Processing gcov (disable by -X gcov)
    Executing gcov (find /home/travis/build/django-polymorphic/django-polymorphic -type f -name '*.gcno'  -exec gcov -pb  {} +)
==> Collecting reports
    + /home/travis/build/django-polymorphic/django-polymorphic/coverage.xml bytes=95006
==> Appending environment variables
    + TRAVIS_PYTHON_NAME
    + TRAVIS_OS_NAME
==> Uploading
    .url https://codecov.io
    .query slug=django-polymorphic%2Fdjango-polymorphic&branch=master&package=py2.0.5&pr=277&job=221934631&commit=a57b067b425a40678238a88cf5cc330cc44b44aa&build=360.16&service=travis
    Pinging Codecov...
    Uploading to S3...
    https://codecov.io/github/django-polymorphic/django-polymorphic/commit/a57b067b425a40678238a88cf5cc330cc44b44aa
```

... and codecov made a comment on this issue.

.travis.yml
meshy: I suspect you may be misunderstanding my intention here.

It would be possible to change this to be simply ` - tox`, but then all the combinations of python and django would run in one travis build. By passing `-e py33-django17` (for example), we can run combination per build, and can make use of Travis features like `allow_failures` and `fast_finish`.

hobarrera: Nope, you actually misunderstood me, heh!

If `TOXENV` is set, tox will run that, so these two are equivalent:

```
tox -e $SOMETHING
TOXENV=$SOMETHING tox
```

As far as the python version goes, if you provide an environment that specifies no python (eg: `django11` instead of `py36-django11`), it'll use the default system python (and just that one!) (which you override in each row in the travis matrix).

So, in practice, running `tox` here is the same as the current command.

Here's an example one and its results on travis in case you want to inspect this a bit more:
* https://github.com/pimutils/todoman/blob/master/.travis.yml
* https://travis-ci.org/pimutils/todoman

tox.ini
meshy: Perhaps there's a better place to put this command, or maybe there's a reason for this problem that I have not spotted?

hobarrera: I haven't seen similar results, so I'm not sure why that is, sorry.

Yeah, it's probably better to put this in travis' `after_script`; it's not part of the test itself (and users running locally won't need this).

.travis.yml
meshy: Ooh, that's fantastic! Thanks!

I had picked the name `TOX_ENV` without realising it had any special meaning... Thanks for the tip!

Apologies for the misunderstanding -- it was indeed me `;)`

I'll make this change soon `:)`

hobarrera: No problem, yeah, I did the same thing at one point, so I get it. Cheers!

meshy:
@vdboor are you up for making this change? If you're willing, then I will extend this approach to add builds for postgres and mysql in another PR. That should help to be sure that #267 is good to merge...

PR #282, opened 2017-05-07
WIP: Run tests against PostgreSQL and MySQL on CI and tox

meshy:
This is not yet ready to review or merge. I've opened a PR so that I can see how TravisCI deals with the builds.

---

I think this is required in order to catch DB specific errors that would otherwise go uncaught.

TODO:

1. This fails all over the place.

    I'll be looking into why.

2. The number of builds (96) is insane.

    As django 2.0 is on the horizon, I'd say it's a good time to drop support for old versions of django and python. I hope to address that particular issue in another PR, but it will reduce the number of builds here to (almost) sane levels.

3. We should consider dropping SQLite on Travis CI.

    That'd cut out 1/3 of the builds immediately, and in my experience it's rarely used in production django environments.


jleclanche:
You can specify multiple environments in the TOXENV variable, such as `TOXENV=django111-postgresql,django111-sqlite`. I would highly recommend doing so.

jleclanche:
(Or alternatively, not do that and maybe instead add separate builds for all supported python versions for each respective Django version!)

meshy:
Thanks for that, I'll bear it in mind once the tests are passing here. That said, I'm hoping that when the number of supported versions of python and django are reduced, and we drop builds for sqlite, that the matrix will be small enough to make that unnecessary.

jleclanche:
Perhaps; it would still reduce build times by quite a bit (and be kinder to Travis).

jleclanche:
@meshy #289 removes compatibility with dinosaur django versions.

meshy:
Fantastic! I'm not sure exactly when I'll get back to this, but it seems wise to wait until that's resolved before continuing.

Issue #257, opened 2017-01-16
Expressions don't work with order_by


hornn:
In Django-1.8 expressions are supported by order_by(), in addition to field names.
For example:
`no_poly = no_poly.order_by(Lower('name').desc())`
is translated to `ORDER BY LOWER("no_poly_table_name"."name") DESC`

However, when the model is polymorphic the same commands fail with parsing exceptions:
```
from django.db.models.functions import Lower

poly = poly.order_by(Lower('name'))
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "lib/python2.7/site-packages/polymorphic/query.py", line 121, in order_by
    new_args = [translate_polymorphic_field_path(self.model, a) for a in args]
  File "lib/python2.7/site-packages/polymorphic/query_translate.py", line 151, in translate_polymorphic_field_path
    classname, sep, pure_field_path = field_path.partition('___')
AttributeError: 'Lower' object has no attribute 'partition'
```
```
from django.db.models import F

poly = poly.order_by('name')  # old syntax works
poly = poly.order_by(F('name'))  # expression fails
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "lib/python2.7/site-packages/polymorphic/query.py", line 121, in order_by
    new_args = [translate_polymorphic_field_path(self.model, a) for a in args]
  File "lib/python2.7/site-packages/polymorphic/query_translate.py", line 151, in translate_polymorphic_field_path
    classname, sep, pure_field_path = field_path.partition('___')
AttributeError: 'F' object has no attribute 'partition'
```

VladBezpalko:
Hi. I had the same problem. In my project I make simple workaround, like this.
`
class FixedPolymorphicQuerySet(PolymorphicQuerySet):

    def order_by(self, *args, **kwargs):
        new_args = [
            translate_polymorphic_field_path(self.model, a)
            if hasattr(a, 'partition') else a
            for a in args
        ]
        return super(PolymorphicQuerySet, self).order_by(*new_args, **kwargs)


class FixedPolymorphicManager(PolymorphicManager):
    queryset_class = FixedPolymorphicQuerySet
`
Dear contributors, what you think about this code? Will this work in all cases?

vdboor:
Fixed in bcb8b0d3a4c6871e1dda1b0321013683a7ed82ba!

PR #288, opened 2017-05-18
Allow PolymorphicQuerySet.order_by to accept django orm expressions

VladBezpalko:
Simple fix for #257


jleclanche:
lgtm

vdboor:
Thanks for the fix!

I've looked into this, and replaced the `if` check a bit in bcb8b0d3a4c6871e1dda1b0321013683a7ed82ba
The `partition()` method is part of the `str` type. I think checking on types is a bit saner here.

PR #289, opened 2017-05-19
Django compatibility cleanups and fixes

### LINKED ISSUE #287: Drop support for old Django versions?
Anyone opposed to dropping support for Django versions older than 1.10? Django 1.11 is now supported in the latest version, 1.10 and 1.11 are the only versions which are still supported upstream, and it looks like there's actually a lot of hacks for these old versions.

This would also mean dropping Python 2.6 support.

I can work on a PR for it if accepted.

jleclanche:
This PR does a few things. In the first two commits:

- We no longer test on versions of Django and Python which are no longer supported.
- Remove all pre-django 1.8 hacks.

Then, in a separate commit, we remove support for Django 1.8 as well. I kept this separate because I may understand the argument to keep Django 1.8 as it is still supported upstream... *however*, Django 1.11 is the new LTS and, given the complexity of django-polymorphic and the amount of hacks dropping 1.8 removes, I would *very highly* recommend dropping 1.8 support as well, and recommend django-polymorphic==1.2 to Django 1.8 users.

As a followup to this, I am also improving compatibility with Django 2.0, to the point that tests now pass on django-dev. I cannot vouch for their reliability, however, especially given that I still see the following warning on 1.11 with -Wall:

```
Using Django version 1.11.1 from /home/adys/tmp/django-polymorphic/.tox/py36-django111/lib/python3.6/site-packages/django
/home/adys/tmp/django-polymorphic/.tox/py36-django111/lib/python3.6/site-packages/django/db/models/base.py:363: RemovedInDjango20Warning: Managers from concrete parents will soon qualify as default managers if they appear before any other managers in the MRO. As a result, 'base_objects' declared on 'polymorphic.PolymorphicModel' will no longer be the default manager for 'polymorphic.MROBase2' in favor of 'objects' declared on 'polymorphic.MROBase1'. You can redeclare 'base_objects' on 'MROBase2' to keep things the way they are or you can switch to the new behavior right away by setting `Meta.manager_inheritance_from_future` to `True`.
  if not opts.managers or cls._requires_legacy_default_manager():
/home/adys/tmp/django-polymorphic/.tox/py36-django111/lib/python3.6/site-packages/django/db/models/base.py:363: RemovedInDjango20Warning: Managers from concrete parents will soon qualify as default managers. As a result, the 'objects' manager won't be created (or recreated) automatically anymore on 'polymorphic.MgrInheritC' and 'mgrB' declared on 'polymorphic.MgrInheritB' will be promoted to default manager. You can declare explicitly `objects = models.Manager()` on 'MgrInheritC' to keep things the way they are or you can switch to the new behavior right away by setting `Meta.manager_inheritance_from_future` to `True`.
  if not opts.managers or cls._requires_legacy_default_manager():
```

Closes #287

### DIFF
--- a/polymorphic/admin/parentadmin.py
+++ b/polymorphic/admin/parentadmin.py
@@ -4,7 +4,6 @@ The parent admin displays the list view of the base model.
 import sys
 import warnings
 
-import django
 from django.conf.urls import url
 from django.contrib import admin
 from django.contrib.admin.helpers import AdminErrorList, AdminForm
@@ -270,31 +269,10 @@ class PolymorphicParentModelAdmin(admin.ModelAdmin):
         if not self._compat_mode:
             return urls
 
-        info = _get_opt(self.model)
-
-        # Patch the change view URL so it's not a big catch-all; allowing all
-        # custom URLs to be added to the end. This is done by adding '/$' to the
-        # end of the regex.  The url needs to be recreated, patching url.regex
-        # is not an option Django 1.4's LocaleRegexProvider changed it.
-        if django.VERSION < (1, 9):
-            # On Django 1.9, the change view URL has been changed from
-            # /<app>/<model>/<pk>/ to /<app>/<model>/<pk>/change/, which is
-            # why we can skip this workaround for Django >= 1.9.
-            new_change_url = url(
-                r'^{0}/$'.format(self.pk_regex),
-                self.admin_site.admin_view(self.change_view),
-                name='{0}_{1}_change'.format(*info)
-            )
-
-            redirect_urls = []
-            for i, oldurl in enumerate(urls):
-                if oldurl.name == new_change_url.name:
-                    urls[i] = new_change_url
-        else:
-            # For Django 1.9, the redirect at the end acts as catch all.
-            # The custom urls need to be inserted before that.
-            redirect_urls = [pat for pat in urls if not pat.name]  # redirect URL has no name.
-            urls = [pat for pat in urls if pat.name]
+        # The redirect at the end acts as catch all.
+        # The custom urls need to be inserted before that.
+        redirect_urls = [pat for pat in urls if not pat.name]  # redirect URL has no name.
+        urls = [pat for pat in urls if pat.name]
 
         # Define the catch-all for custom views
         custom_urls = [
@@ -421,7 +399,3 @@ class PolymorphicParentModelAdmin(admin.ModelAdmin):
             "admin/%s/change_list.html" % base_app_label,
             "admin/change_list.html"
         ]
-
-
-def _get_opt(model):
-    return model._meta.app_label, model._meta.model_name
--- a/polymorphic/admin/parentadmin.py
+++ b/polymorphic/admin/parentadmin.py
@@ -10,9 +10,9 @@ from django.contrib.admin.helpers import AdminErrorList, AdminForm
 from django.contrib.admin.templatetags.admin_urls import add_preserved_filters
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import PermissionDenied
-from django.core.urlresolvers import RegexURLResolver
 from django.http import Http404, HttpResponseRedirect
 from django.template.response import TemplateResponse
+from django.urls import RegexURLResolver
 from django.utils.encoding import force_text
 from django.utils.http import urlencode
 from django.utils.safestring import mark_safe
--- a/polymorphic/admin/parentadmin.py
+++ b/polymorphic/admin/parentadmin.py
@@ -75,38 +75,8 @@ class PolymorphicParentModelAdmin(admin.ModelAdmin):
         if self._is_setup:
             return
 
-        # By not having this in __init__() there is less stress on import dependencies as well,
-        # considering an advanced use cases where a plugin system scans for the child models.
-        child_models = self.get_child_models()
-        # Check if get_child_models() returns an iterable of models (new format) or an iterable
-        # of (Model, Admin) (legacy format). When iterable is empty, assume the new format.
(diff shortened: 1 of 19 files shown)


jleclanche:
@vdboor Any initial thoughts? :)

vdboor:
Hi! I love your work on cleaning up polymorphic. This would be a great start for a 2.0 release.
Anything below 1.8 is good to go, with me too! I'm not really enthusiastic about removing Django 1.8 support too, as that makes upgrading from one LTS to the other a lot harder. 

We've always been very hesitant to remove old code, and this takes it to the other extreme. What are your thoughts on this?

jleclanche:
I think it's 100% safe to remove anything that is no longer supported upstream. I do it pretty religiously in all my libraries.

Regarding dropping 1.8 compatibility, it's of course trickier since it's still supported until next April. However, 1.8 is supported on a security-only basis. It is imho completely reasonable to drop support for it in a point-release, and, if necessary, do minor security releases on the older release should the need arise. It's also a lot simpler to deal with than a bunch of compat hacks.

Additionally, polymorphic is imho a case of a fairly complex library which deals with Django's inner workings quite a lot. Django has changed a bunch since 1.8, and supporting it in the most recent release increases surface area a lot. You can see it in the commit stats: `+236 insertions, -912 deletions`.

Finally, Django themselves recommended that once Django 2.0 is out, libraries drop support for everything except Django 1.11 and 2.0. It's not out yet but it's a decent idea to prepare for it.

I would recommend landing this PR, as well as a cleaned up #282 (I can help with it), and do a new major release shortly after (1.3 or 2.0, whatever semver you think is worth bumping for dropping so many versions of Django at once). Put the previous major branch in security-only maintenance.

jleclanche:
@vdboor How are we doing?

jleclanche:
@vdboor ping

jleclanche:
Do not merge. There are currently issues with setuptools' latest version.

jleclanche:
@vdboor ping, this is once again good to land

jleclanche:
This PR now requires #291; please land the latter before this one :)

polymorphic/admin/parentadmin.py
```diff
@@ -334,9 +293,9 @@ def subclass_view(self, request, path):
             try:
                 pos = path.find('/')
                 if pos == -1:
-                    object_id = long(path)
+                    object_id = int(path)
```
jleclanche: It won't, long/int in py2.7 is handled the same way.

jleclanche:
> The part that I'm still hesitant about is removing 1.8 support already. In fact for similar reasons that Django devs recommend to support 1.11 and 2.0: to make upgrading easy. While it likely doesn't matter for many libraries, polymorphic has been a cornerstone for a lot of projects, so I've kept this tradition of long compatibility when I took maintainership. However, I also see that it makes entry into the project very hard now.

I'm with you on this, however for anyone using 1.8 it's still possible to remain on an older version of Polymorphic, then upgrade to the non-1.8 one later.

You could make a release, then land all this, then make another release. It's what I'd do if I wanted to offer an upgrade path.

polymorphic/admin/parentadmin.py
jleclanche: There's also no reason to use `long()` here really.

jleclanche:
@vdboor What do you think? And is a release possible soon?

meshy:
I think I agree with @vdboor on the django 1.8 thing. I'm keen to see it dropped too, but splitting it into another PR would make merging this an easier pill to swallow, I imagine.

jleclanche:
The PR is already split that way like i said in the original comment. The first two comments only drop pre-1.8 stuff. Everything else is separate cleanups.

meshy:
Oh, I understand that, and I applaud the fact that you've kept the commits clean.

My suggestion was to split the 1.8-and-up stuff into another PR that can be merged once django drops support for it, and then this could probably be merged now.

jleclanche:
Well, I can definitely split up the commits that can just land as is if you think that can help get things moving, but I'd like to strongly reiterate my recommendation to drop django 1.8 support earlier than Django upstream for the reasons I outlined in https://github.com/django-polymorphic/django-polymorphic/pull/289#issuecomment-303065664

jleclanche:
This PR now depends on #298. #298 can be landed for this release, this PR for the next.

vdboor:
Thanks for the change; I've left a comment on #298 regarding failed tests. Other then that, everything looks fine to merge #298 !  :-D

jleclanche:
This is ready to land. @vdboor Am I good to merge this?

polymorphic/tests/test_utils.py
```diff
@@ -28,9 +28,13 @@ def test_reset_polymorphic_ctype(self):
...
+
+        field_reprs = [
+            "<Model2A: id 1, field1 (CharField)>",
+            "<Model2D: id 2, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField)>",
+            "<Model2B: id 3, field1 (CharField), field2 (CharField)>",
+            "<Model2B: id 4, field1 (CharField), field2 (CharField)>",
+        ]
+
+        for f, f_repr in zip(Model2A.objects.order_by("pk"), field_reprs):
+            self.assertEqual(qrepr(f), f_repr)
```
hobarrera: [`assertQuerysetEqual`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.TransactionTestCase.assertQuerysetEqual) might actually be what you're looking for.

jleclanche:
Any objections to merging this? It's ready now, all tests passing etc. This *does* drop Django 1.8 compatibility for next release which makes polymorphic 1.4 Django 1.10+ only.

hobarrera:
Can't think of any reason not to merge. 🙂

PR #293, opened 2017-06-05
Fixed PO files which resulted in SyntaxError when parsed by polib

NeveHanter:
We're using django-rosetta which uses polib to parse PO files and it fails with django-polymorphic's en/fr PO files resulting in SyntaxError.

This applies to both `#msgid` and `#msgstr` entries and problem lies that comments in PO files should have space after `#` character, otherwise they're handled as unknown token, thus SyntaxError.

This PR fixes the problem.


jleclanche:
Looks good to merge.

vdboor:
Merged!

PR #298, opened 2017-06-24
Drop Django < 1.8 compatibility

jleclanche:
This is a subset of #289 which is safe to land for the next release, per the comments.


polymorphic/admin/parentadmin.py
```diff
@@ -334,9 +315,9 @@ def subclass_view(self, request, path):
             try:
                 pos = path.find('/')
                 if pos == -1:
-                    object_id = long(path)
+                    object_id = int(path)
```
hobarrera: Are you entirely sure this won't affect Python 2.7 negatively?

jleclanche: Yes, positive.

vdboor:
Thanks for providing this pull request!

When I have some time, I'll fix the remaining issue (or someone else should beat me first). I think the the `qrepr()` is not completely compatible with 1.8, which breaks tests, because everything else looks fine!

jleclanche:
@vdboor any luck figuring it out?

polymorphic/admin/parentadmin.py
vdboor: I'm not so sure, as Django's `BigIntegerField` can produce longs on Python 2. I've reverted this in 67dd5c0a0dd33537c1deb946dbae0c8e98fbecb2

vdboor:
Thanks a lot for all your work @jleclanche! :-D This work is merged now, and it should make maintenance a lot easier for new developers. 👍

jleclanche:
Excellent! Release soon? :)

jleclanche:
@vdboor poke poke, is the release still blocked on something? can I help with anything?

Issue #51, opened 2013-07-29
DoesNotExist: ContentType matching query does not exist. Lookup parameters were {'pk': None}


taxido:
Django 1.5.1
After south migration.

``` python
In [1]: Item.objects.all()
Out[1]: ---------------------------------------------------------------------------
DoesNotExist                              Traceback (most recent call last)
<ipython-input-1-f68cadf55917> in <module>()
----> 1 Item.objects.all()

/home/vagrant/env/local/lib/python2.7/site-packages/IPython/core/displayhook.pyc in __call__(self, result)
    236             self.start_displayhook()
    237             self.write_output_prompt()
--> 238             format_dict = self.compute_format_data(result)
    239             self.write_format_data(format_dict)
    240             self.update_user_ns(result)

/home/vagrant/env/local/lib/python2.7/site-packages/IPython/core/displayhook.pyc in compute_format_data(self, result)
    148             MIME type representation of the object.
    149         """
--> 150         return self.shell.display_formatter.format(result)
    151 
    152     def write_format_data(self, format_dict):

/home/vagrant/env/local/lib/python2.7/site-packages/IPython/core/formatters.pyc in format(self, obj, include, exclude)
    124                     continue
    125             try:
--> 126                 data = formatter(obj)
    127             except:
    128                 # FIXME: log the exception

/home/vagrant/env/local/lib/python2.7/site-packages/IPython/core/formatters.pyc in __call__(self, obj)
    445                 type_pprinters=self.type_printers,
    446                 deferred_pprinters=self.deferred_printers)
--> 447             printer.pretty(obj)
    448             printer.flush()
    449             return stream.getvalue()

/home/vagrant/env/local/lib/python2.7/site-packages/IPython/lib/pretty.pyc in pretty(self, obj)
    358                             if callable(meth):
    359                                 return meth(obj, self, cycle)
--> 360             return _default_pprint(obj, self, cycle)
    361         finally:
    362             self.end_group()

/home/vagrant/env/local/lib/python2.7/site-packages/IPython/lib/pretty.pyc in _default_pprint(obj, p, cycle)
    478     if getattr(klass, '__repr__', None) not in _baseclass_reprs:
    479         # A user-provided repr.
--> 480         p.text(repr(obj))
    481         return
    482     p.begin_group(1, '<')

/home/vagrant/env/local/lib/python2.7/site-packages/polymorphic/query.pyc in __repr__(self, *args, **kwargs)
[...]

PR #304, opened 2017-07-17
fix the content type if it's missing (from migrations)

pwag42:
From https://github.com/django-polymorphic/django-polymorphic/issues/51

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -363,7 +363,11 @@ class PolymorphicQuerySet(QuerySet):
                     results[base_object.pk] = base_object
 
                 else:
-                    real_concrete_class = base_object.get_real_instance_class()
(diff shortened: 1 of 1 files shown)


vdboor:
Sorry, I understand you'd like to fix this - but the given solution is not the right one. Mainly:
- the exception clause is too broad.
- fetching results from a database now changes the database.

Instead, polymorphic now provides a `polymorphic.utils.reset_polymorphic_ctype` function that does a much better job. I'll try to create a management command for this, and make it even more intuitive to work with (e.g. just specify the base model, and let everything be sorted out)

pwag42:
Sounds good to me! I just threw that in because it worked for me and missing content types seemed to be a common problem.

________________________________

Issue #270, opened 2017-03-02
PolymorphicModel fails if swapped with another Model in Django 1.10


GheloAce:
I developed a reusable app called `alpha` that provides the `Alpha` model (that is expected to get inherited by another model `AnotherAlpha`) using `django-polymorphic` and `swapper`.  Then created the `beta` app that inherits `alpha`'s `AlphaAbstract` so that users can decide to use a different Model instead of the default.  After swapping to `beta`'s implementation of `Alpha` i get this error:

```
  [...]
  File "/home/ghelo/PycharmProjects/capricorn/PolySwapper/alpha/models.py", line 10, in <module>
    class Alpha(AlphaAbstract):
  File "/home/ghelo/PycharmProjects/capricorn/lib/python3.6/site-packages/polymorphic/base.py", line 99, in __new__
    self.validate_model_manager(new_class.objects, model_name, 'objects')
  File "/home/ghelo/PycharmProjects/capricorn/lib/python3.6/site-packages/django/db/models/manager.py", line 198, in __get__
    cls._meta.swapped,

AttributeError: Manager isn't available; 'alpha.Alpha' has been swapped for 'beta.Beta'
```

---------------------
## Replication

**1.** Run these commands:
```bash
pip install Django
django-admin startproject PolySwapper
cd PolySwapper
django-admin startapp alpha
django-admin startapp beta
pip install django-polymorphic swapper
```
As of reporting this issue the package versions are:
```
Django==1.10.6
swapper==1.0.0
django-polymorphic==1.1
```
**2.** Create models Alpha and Beta
```python
# alpha/models.py
import swapper
from polymorphic.models import PolymorphicModel
class AlphaAbstract(PolymorphicModel):
    class Meta:
        abstract = True
class Alpha(AlphaAbstract):
    class Meta:
        swappable = swapper.swappable_setting('alpha', "Alpha")
class AnotherAlpha(Alpha):
    pass

# beta/models.py
from alpha.models import AlphaAbstract
class Beta(AlphaAbstract):
    pass
```
**3.** Run `./manage.py makemigrations`
**4.** Add `alpha` and `beta` to `settings.py`:
```python
INSTALLED_APPS = [
    ...
    'alpha',
    'beta',
]
```
**5.** Append the following to `settings.py` like so:
```python
ALPHA_ALPHA_MODEL = "beta.Beta"
```

**6.** Run `./manage.py shell`  (or any command) and you'll get the error:
```
[...]

PR #307, opened 2017-07-25
Swapped polymorphic models don't work on Django>=1.10

skirsdeda:
This is so because polymorphic tries to validate manager on a swapped model while such manager is inaccessible (described in detail in #270)

### DIFF
--- a/polymorphic/base.py
+++ b/polymorphic/base.py
@@ -94,11 +94,12 @@ class PolymorphicModelBase(ModelBase):
                 new_class._default_manager = user_manager._copy_to_model(new_class)
                 new_class._default_manager._inherited = False   # the default mgr was defined by the user, not inherited
 
-        # validate resulting default manager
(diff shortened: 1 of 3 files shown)


skirsdeda:
Test on Django 1.10 and 1.11 have failed as expected.

skirsdeda:
And with the fix the new tests pass :)

hvdklauw:
Hmm kinda sad that this fix is only in the 2.0 release which doesn't support django 1.10, while this issue is very much present in 1.10, now I'll have to skip a whole test step in my path to upgrade a project from django 1.8 to 1.11 because there is no version that works

vdboor:
@hvdklauw @skirsdeda  django-polymorphic 1.3.1 is released with these changes included. Since I got more of these requests,  @skirsdeda inspired me to do a 1.3.1 release. It contains almost all bugfixes from the 2.0 release

skirsdeda:
@vdboor Great!

PR #311, opened 2017-08-30
fix error with .defer and child models that use the same parent

trbs:
When using .defer on a PolymorphicQuerySet with multiple childs that
subclass from the same polymorphic parent model yield an error like:

```
>>> Base.objects.defer('ModelY___field_y')
...
FieldDoesNotExist: ModelX has no field named 'field_y'
```


vdboor:
Thanks for the clear unit test, and fix!

PR #312, opened 2017-09-06
Fix failing tests with django/master

### LINKED ISSUE #310: Failing test on django master // python 3.6
https://travis-ci.org/django-polymorphic/django-polymorphic/jobs/262018540

I don't know why this happens, but should look at it at some point.

hobarrera:
Several tests were failing with django master. This is because, apparently, `id`s don't match those expected in tests, and other test-related details.

Replace all test comparisons with `assertQuerysetEqual`, which is safer and cleaner for the sort of comparison we're doing.

Fixes #310

### DIFF
--- a/polymorphic/tests/test_orm.py
+++ b/polymorphic/tests/test_orm.py
@@ -2,14 +2,14 @@ import re
 
 import django
 from django.db.models import Case, Count, Q, When
-from django.test import TestCase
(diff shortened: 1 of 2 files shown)


hobarrera:
Thanks!

jleclanche:
No, thank you!

PR #318, opened 2017-10-29
Add django 2.0 support

meshy:
This fixes a few minor issues to add support for django 2.0(b1).

I have:

- Removed django 1.10 from the test suite in the process, ~but have not yet dug out compat fixes that were added for it~.
- Removed three models from the tests that were never referenced in the tests.
- Not yet tested this manually.

There is one remaining unaddressed `RemovedInDjango20Warning`, but I've had some frustration with it. See https://code.djangoproject.com/ticket/28750

TODO:

- [x] Remove django 1.10 compat fixes.
- [x] Chase up related django ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,8 @@ language: python
 python: "3.6"
 
 env:
-  - TOXENV=py27-django110
   - TOXENV=py27-django111
-  - TOXENV=py34-django110
   - TOXENV=py34-django111
-  - TOXENV=py35-django110
   - TOXENV=py35-django111
   - TOXENV=py35-djangomaster
   - TOXENV=py36-django111
@@ -25,14 +22,10 @@ matrix:
   fast_finish: true
   include:
     - python: "3.5"
-      env: TOXENV=py35-django110
-    - python: "3.5"
       env: TOXENV=py35-django111
     - python: "3.5"
       env: TOXENV=py35-djangomaster
   exclude:
-    - python: "3.6"
-      env: TOXENV=py35-django110
     - python: "3.6"
       env: TOXENV=py35-django111
     - python: "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,16 @@ python: "3.6"
 env:
   - TOXENV=py27-django111
   - TOXENV=py34-django111
+  - TOXENV=py34-django200
   - TOXENV=py35-django111
+  - TOXENV=py35-django200
   - TOXENV=py35-djangomaster
   - TOXENV=py36-django111
+  - TOXENV=py36-django200
   - TOXENV=py36-djangomaster
   # XXX: Use a matrix to build these?
   - TOXENV=py36-django111-postgres DB=postgres
+  - TOXENV=py36-django200-postgres DB=postgres
   - TOXENV=py36-djangomaster-postgres DB=postgres
 
 services:
@@ -21,13 +25,21 @@ services:
 matrix:
   fast_finish: true
   include:
+    - python: "2.7"
+      env: TOXENV=py27-django111
     - python: "3.5"
       env: TOXENV=py35-django111
+    - python: "3.5"
+      env: TOXENV=py35-django200
     - python: "3.5"
       env: TOXENV=py35-djangomaster
   exclude:
     - python: "3.6"
+      env: TOXENV=py27-django111
+    - python: "3.6"
       env: TOXENV=py35-django111
+    - python: "3.6"
+      env: TOXENV=py35-django200
     - python: "3.6"
       env: TOXENV=py35-djangomaster
   allow_failures:
--- a/polymorphic/formsets/utils.py
+++ b/polymorphic/formsets/utils.py
@@ -1,11 +1,17 @@
 """
 Internal utils
 """
+import django
 
 
 def add_media(dest, media):
     """
     Optimized version of django.forms.Media.__add__() that doesn't create new objects.
+
+    Only required for Django < 2.0
     """
-    dest.add_css(media._css)
-    dest.add_js(media._js)
+    if django.VERSION >= (2, 0):
+        dest += media
+    else:
+        dest.add_css(media._css)
+        dest.add_js(media._js)
--- a/polymorphic/models.py
+++ b/polymorphic/models.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
+from django.db.models.fields.related import ReverseOneToOneDescriptor, ForwardManyToOneDescriptor
 from django.db.utils import DEFAULT_DB_ALIAS
 from django.utils import six
 
@@ -178,14 +179,6 @@ class PolymorphicModel(six.with_metaclass(PolymorphicModelBase, models.Model)):
 
         subclasses_and_superclasses_accessors = self._get_inheritance_relation_fields_and_models()
 
-        try:
-            from django.db.models.fields.related import ReverseOneToOneDescriptor, ForwardManyToOneDescriptor
-        except ImportError:
-            # django < 1.9
-            from django.db.models.fields.related import (
-                SingleRelatedObjectDescriptor as ReverseOneToOneDescriptor,
-                ReverseSingleRelatedObjectDescriptor as ForwardManyToOneDescriptor,
-            )
         for name, model in subclasses_and_superclasses_accessors.items():
             # Here be dragons.
             orig_accessor = getattr(self.__class__, name, None)
--- a/polymorphic/tests/test_orm.py
+++ b/polymorphic/tests/test_orm.py
@@ -228,14 +228,14 @@ class PolymorphicTests(TransactionTestCase):
         objects_deferred = Model2A.objects.defer('field1')
 
         self.assertNotIn('field1', objects_deferred[0].__dict__, 'field1 was not deferred (using defer())')
-        self.assertEqual(repr(objects_deferred[0]),
-                         '<Model2A: id 1, field1 (CharField), deferred[field1]>')
-        self.assertEqual(repr(objects_deferred[1]),
-                         '<Model2B: id 2, field1 (CharField), field2 (CharField), deferred[field1]>')
-        self.assertEqual(repr(objects_deferred[2]),
-                         '<Model2C: id 3, field1 (CharField), field2 (CharField), field3 (CharField), deferred[field1]>')
-        self.assertEqual(repr(objects_deferred[3]),
-                         '<Model2D: id 4, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField), deferred[field1]>')
+        self.assertRegex(repr(objects_deferred[0]),
+                         '<Model2A: id \d+, field1 \(CharField\), deferred\[field1\]>')
+        self.assertRegex(repr(objects_deferred[1]),
+                         '<Model2B: id \d+, field1 \(CharField\), field2 \(CharField\), deferred\[field1\]>')
+        self.assertRegex(repr(objects_deferred[2]),
+                         '<Model2C: id \d+, field1 \(CharField\), field2 \(CharField\), field3 \(CharField\), deferred\[field1\]>')
+        self.assertRegex(repr(objects_deferred[3]),
+                         '<Model2D: id \d+, field1 \(CharField\), field2 \(CharField\), field3 \(CharField\), field4 \(CharField\), deferred\[field1\]>')
 
         objects_only = Model2A.objects.only('pk', 'polymorphic_ctype', 'field1')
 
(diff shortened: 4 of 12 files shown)


tox.ini
```diff
@@ -14,10 +14,11 @@ setenv =
 deps =
 	coverage
 	dj-database-url
-	django110: Django >= 1.10, < 1.11
 	django111: Django >= 1.11, < 2.0
+	django200: Django ~= 2.0b1
 	djangomaster: https://github.com/django/django/archive/master.tar.gz
 	postgres: psycopg2
+pip_pre = True
```
meshy: This will not be required once django 2.0 comes out. Also, the line above will need to change to `Django ~= 2.0.0`

meshy:
I'm not able to reproduce the python 2.7 failure here locally, unfortunately.

polymorphic/compat.py
```diff
@@ -0,0 +1,10 @@
+import django
+
+DJANGO_GTE_2 = django.VERSION >= (2, 0)
```
vdboor: I'd place `django.VERSION >= (2, 0)` or even better: `django.VERSION < (2, 0)` directly inline in the code

polymorphic/compat.py
```diff
@@ -0,0 +1,10 @@
...
+
+try:
+    # Django 2.0+
+    from django.urls import URLResolver
+except ImportError:
+    # Django < 2.0
+    from django.urls import RegexURLResolver as URLResolver
```
vdboor: There is only a single place where this import is needed. Does that quality for a separate file to do this?

tox.ini
vdboor: This can be updated to Django 2.0rc1

polymorphic/models.py
```diff
@@ -50,14 +52,14 @@ class PolymorphicModel(six.with_metaclass(PolymorphicModelBase, models.Model)):
...
     polymorphic_internal_model_fields = ['polymorphic_ctype']
 
-    # Note that Django 1.5 removes these managers because the model is abstract.
-    # They are pretended to be there by the metaclass in PolymorphicModelBase.get_inherited_managers()
     objects = PolymorphicManager()
     base_objects = models.Manager()
 
     class Meta:
         abstract = True
         base_manager_name = "objects"
```
vdboor: do we need a `default_manager_name` too?

polymorphic/models.py
```diff
@@ -50,14 +52,14 @@ class PolymorphicModel(six.with_metaclass(PolymorphicModelBase, models.Model)):
...
+        if not DJANGO_GTE_2:
```
vdboor: `if django.VERSION < (2, 0)` would be clean here too.

polymorphic/formsets/utils.py
```diff
@@ -1,11 +1,17 @@
...
 def add_media(dest, media):
     """
     Optimized version of django.forms.Media.__add__() that doesn't create new objects.
+
+    Only required for Django < 2.0
     """
-    dest.add_css(media._css)
-    dest.add_js(media._js)
+    if DJANGO_GTE_2:
+        dest += media
```
vdboor: ⚠️  This no longer updates the `dest` object so it will break. Returning `dest` and updating `polymorphic/admin/inlines.py`, `polymorphic/formsets/models.py` would be better.

tony:
I also have a branch I've started on my own at https://github.com/develtech/django-polymorphic/tree/django-2.0. 

I'm going to consider closing #327 and #328 in favor of moving this one along.

tox.ini
tony: We can now:

- remove ``pip_pre`` line
- ``django200: Django >= 2.0, < 2.1``

.travis.yml
```diff
@@ -24,17 +25,21 @@ services:
...
       env: TOXENV=py35-djangomaster
   exclude:
     - python: "3.6"
-      env: TOXENV=py35-django110
+      env: TOXENV=py27-django111
     - python: "3.6"
       env: TOXENV=py35-django111
+    - python: "3.6"
+      env: TOXENV=py35-django200
     - python: "3.6"
```
tony: LGTM

tony:
@meshy Can you swing around to removing the compat module/inlining version checks?

tony:
https://travis-ci.org/django-polymorphic/django-polymorphic/jobs/300629621

I don't know where that is coming from. Here's the recreation:

```
git remote add meshy https://github.com/meshy/django-polymorphic.git
git fetch meshy
git checkout meshy/django-2.0
tox
```

Everything greenlights except for Python 2.7 + Django 1.11

```
tox -e py27-django111
```

Response:

```
LOB sdist-make: /Users/me/work/python/django-polymorphic/setup.py
py27-django111 inst-nodeps: /Users/me/work/python/django-polymorphic/.tox/dist/django-polymorphic-1.3.zip
py27-django111 installed: coverage==4.4.2,dj-database-url==0.4.2,Django==1.11.8,django-polymorphic==1.3,pytz==2017.3
py27-django111 runtests: PYTHONHASHSEED='1573255718'
py27-django111 runtests: commands[0] | coverage run --source polymorphic runtests.py
Using Python version 2.7.1 from /Users/me/work/python/django-polymorphic/.tox/py27-django111/bin/python2.7
Using Django version 1.11.8 from /Users/me/work/python/django-polymorphic/.tox/py27-django111/lib/python2.7/site-packages/django
Creating test database for alias 'default'...
Creating test database for alias 'secondary'...
System check identified no issues (0 silenced).
=====================================================================
FAIL: test_defer_fields (polymorphic.tests.test_orm.PolymorphicTests)
----------------------------------------------------------------------
  File "/Users/me/work/python/django-polymorphic/polymorphic/tests/test_orm.py", line 227, in test_defer_fields
    '<Model2A: id 1, field1 (CharField), deferred[field1]>')
AssertionError: '<Model2A: id 5, field1 (CharField), deferred[field1]>' != '<Model2A: id 1, field1 (CharField), deferred[field1]>'

----------------------------------------------------------------------
Ran 57 tests in 9.678s

FAILED (failures=1)
Destroying test database for alias 'default'...
Destroying test database for alias 'secondary'...
ERROR: InvocationError: '/Users/me/work/python/django-polymorphic/.tox/py27-django111/bin/coverage run --source polymorphic runtests.py'
```

tony:
Is the ID of the model an important detail to test against when using defer()?

https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.defer

It *is* strange that id isn't acting as expected. I got no idea where that's coming from.

Flipping around sqlite addresses:

```
diff --git a/runtests.py b/runtests.py
index bb3074d..810a58e 100755
--- a/runtests.py
+++ b/runtests.py
@@ -19,14 +19,14 @@ if not settings.configured:
     settings.configure(
         DEBUG=False,
         DATABASES={
-            'default': dj_database_url.config(
-                env='PRIMARY_DATABASE',
-                default='sqlite://:memory:',
-            ),
-            'secondary': dj_database_url.config(
-                env='SECONDARY_DATABASE',
-                default='sqlite://:memory:',
-            ),
+            'default': {
+                'ENGINE': 'django.db.backends.sqlite3',
+                'NAME': 'file:mydatabase2.db',
+            },
+            'secondary': {
+                'ENGINE': 'django.db.backends.sqlite3',
+                'NAME': 'file:mydatabase44.db',
+            },
         },
         TEST_RUNNER="django.test.runner.DiscoverRunner",
         INSTALLED_APPS=(
```

Changed nothing. It doesn't seem to be anything related to secondary db (same result happens if second db removed from settings).

I also ruled out it being due to other tests inside the same class (which shouldn't matter anyway if its in a ``TransactionTestCase``)

I'm not sure if the ID is as important as the field showing up deferred.

tony:
OK, I've ruled out the above error not being related to this PR, it started here 80b4f2b.

So other than the bikeshed stuff (removing the compat file, updating versions and the ``pip_pre``) this is good for another look.

polymorphic/compat.py
meshy: Sure ok. WRT this and your other comment in this `compat` file, please excuse me, I was working on habit here. I've found that it's helpful to have a single point of reference for compatibility hacks, as it makes maintenance easier down the road. I probably shouldn't have introduced it here without running it past you first though :)

I'll move this inline, as you have suggested.

polymorphic/formsets/utils.py
meshy: Are you 100% sure this fails? I thought I had tested this quite thoroughly.

If you're sure you'd like this to change, please confirm, and I'll get it done.

meshy:
@tony thank you for the investigation into the failing test. It's a great relief to know that the cause does not stem from this PR.

meshy:
I've made most of the requested changes here, and have rebased the commits. Still waiting on a confirmation on the changes requested to `polymorphic/formsets/utils.py` before making a significant change there.

I realise it's taken me quite a while to get back to this one -- thank you for your patience!

tony:
I closed my #327 and #328 because they duplicate the changes here.

meshy:
This is blocked because of failing tests. I believe the issue is in the `master` branch. See https://github.com/django-polymorphic/django-polymorphic/issues/331.

vdboor:
The fix in #322 is merged! Sorry for the delay. Would this clear the way for Django 2.0 now?

meshy:
Thanks for merging that. I'm still getting that one test failure, but I should be able to investigate it now. It's frustrating that I'm not getting the failure in local tests; only on Travis.

meshy:
@tony did you get to the bottom of what might be causing the issue?

tony:
(I have a lot on my plate ATM, and am not using django-polymorphic in my sites anymore.)

I think that deferred issue started happening in 80b4f2b.  I don't think it's related to this PR.

I notice that in https://github.com/django-polymorphic/django-polymorphic/blob/80b4f2bb41423d6131e2e39ef5c6b212ce5a7b56/polymorphic/tests/test_orm.py#L232, the asserts weren't updated to be wrapped in repr.

tony:
See #332 for the last test failure. But I think this PR is ok.

polymorphic/tests/test_orm.py
```diff
@@ -228,14 +228,14 @@ def test_defer_fields(self):
...
         self.assertNotIn('field1', objects_deferred[0].__dict__, 'field1 was not deferred (using defer())')
-        self.assertEqual(repr(objects_deferred[0]),
-                         '<Model2A: id 1, field1 (CharField), deferred[field1]>')
-        self.assertEqual(repr(objects_deferred[1]),
-                         '<Model2B: id 2, field1 (CharField), field2 (CharField), deferred[field1]>')
-        self.assertEqual(repr(objects_deferred[2]),
-                         '<Model2C: id 3, field1 (CharField), field2 (CharField), field3 (CharField), deferred[field1]>')
-        self.assertEqual(repr(objects_deferred[3]),
-                         '<Model2D: id 4, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField), deferred[field1]>')
+        self.assertRegex(repr(objects_deferred[0]),
```
WhyNotHugo: Honestly, this could really be an `assertQuerySetEqual`, which is more future-proof, and quite cleaner.

Not a blocker for this specific PR though.

meshy: I'm not really sure how I would make that change. Thanks for not blocking on it!

meshy:
Tests now appear to be passing! `\o/`

polymorphic/formsets/utils.py
```diff
@@ -1,11 +1,17 @@
...
 def add_media(dest, media):
     """
     Optimized version of django.forms.Media.__add__() that doesn't create new objects.
+
+    Only required for Django < 2.0
     """
-    dest.add_css(media._css)
-    dest.add_js(media._js)
+    if django.VERSION >= (2, 0):
+        dest += media
```
vdboor: This will break. The Django code for `dest +=` does the following:

```
    def __add__(self, other):
        combined = Media()
        combined._js = self.merge(self._js, other._js)
        combined._css = {
            medium: self.merge(self._css.get(medium, []), other._css.get(medium, []))
            for medium in self._css.keys() | other._css.keys()
        }
        return combined
```

Thus, the `dest` object that was passed to the function is not changed here, while that was the whole point of the `add_media` performance boost.

vdboor: However, that was easy to fix, so I've done that in 9d651a1cdb92d7d8ba039fce97a11de085b54990 :-)

meshy: It took me far too long to understand why it didn't work. Thanks for fixing that! `:)`

vdboor:
I'm super happy to see all the collaboration to get this PR working. If one of you guys like to have a commit-bit for this project, you're more then welcome! Just mention it here!

Having more people to take up work only seems healthy to me :-)

mblayman:
Kudos to you all for getting Django 2.0 support added!

I'm the maintainer for Django REST Framework JSON API (https://github.com/django-json-api/django-rest-framework-json-api), and its test suite relies on django-polymorphic for some polymorphic capable serializers that were added. I'm eagerly awaiting this project's next release so I can push out a Django 2.0 compatible release of DJA. 😄 Nice work!

PR #321, opened 2017-11-02
Do not use deprecated and removed declared_fieldsets attribute in admin

skirsdeda:
`ModelAdmin.declared_fieldsets` property was removed in Django 1.9. Checking `fieldsets` and `fields` is equivalent and will actually make that `if` do it's work.


vdboor:
thanks!

tony:
This seems to have caused some breakage #331

skirsdeda:
@tony, that is already fixed in my new PR #322 (which also covers the typo you fixed in your PR).

tony:
@skirsdeda Thank you! I didn't see that!

That'll help move things along with getting 2.0 working.

PR #322, opened 2017-11-20
Admin fieldsets fix (another attempt)

skirsdeda:
Sorry for this mess. The checks for `fields` and `fieldsets` were completely wrong in the previous PR (which you fixed, albeit with a typo 'or or').
Also, the "prevent infinite recursion" lines that I deleted are actually necessary. So I restored that with correct fields/fieldsets check now.

### DIFF
--- a/polymorphic/admin/childadmin.py
+++ b/polymorphic/admin/childadmin.py
@@ -62,6 +62,10 @@ class PolymorphicChildModelAdmin(admin.ModelAdmin):
         # If the derived class sets the model explicitly, respect that setting.
         kwargs.setdefault('form', self.base_form or self.form)
 
+        # prevent infinite recursion when this is called from get_subclass_fields
+        if not hasattr(self, 'fieldsets') and not hasattr(self, 'fields'):
+            kwargs.setdefault('fields', None)
+
         return super(PolymorphicChildModelAdmin, self).get_form(request, obj, **kwargs)
 
     def get_model_perms(self, request):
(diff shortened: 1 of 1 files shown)


tony:
This would fix #331

tony:
```
./runtests.py
Using Python version 3.6.3 from /Users/me/work/python/django-polymorphic/.venv/bin/python
Using Django version 1.11.4 from /Users/me/work/python/django-polymorphic/.venv/lib/python3.6/site-packages/django
Creating test database for alias 'default'...
Creating test database for alias 'secondary'...
System check identified no issues (0 silenced).
..........................................................
----------------------------------------------------------------------
Ran 58 tests in 5.991s

OK
Destroying test database for alias 'default'...
Destroying test database for alias 'secondary'...
```

I don't want to have stuff merged in before the time is ripe, but I do want travis to have greenlights while I chomp away at 2.0 support. The changes for that have to support 1.11 at least (and 1.10 if it's not agreed upon in #326).

tony:
I'm working on getting a 2.0 branch cleared up / QA'd at https://github.com/develtech/django-polymorphic/tree/django-2.0. I preliminarily merged this PR to work against for the time being.

tony:
I think the breakage caused by 80b4f2b (#332) can be safely ignored in this PR also.

tony:
This PR is blocking CI on other PR's. #318

tony:
@vdboor can you commit to master to remove the "or or" part out to see if that gets CI greenlighting on #318 

Even though tests pass on this, I don't use custom fields on django polymorphic.  Is there a way you could test this behavior @skirsdeda ?

ShaheedHaque:
@tony why would this change affect or be affected by the type of a form field? (I'm keen to see #318 move forward, and this looks like it would fix the CI issue blocking it).

vdboor:
Thanks guys for this fix! Sorry for the delay on my side.

Issue #325, opened 2017-12-06
Update travis for Django 2.0


tony:
Related #324

vdboor:
Fixed, and 2.0 is released!

Issue #324, opened 2017-12-03
django 2.0 needs support (urls)


tony:
```
  File "/Users/me/.local/share/virtualenvs/develtech-oQr_FNyw/lib/python3.6/site-packages/polymorphic/admin/__init__.py", line 8, in <module>
    from .parentadmin import PolymorphicParentModelAdmin
  File "/Users/me/.local/share/virtualenvs/develtech-oQr_FNyw/lib/python3.6/site-packages/polymorphic/admin/parentadmin.py", line 14, in <module>
    from django.core.urlresolvers import RegexURLResolver
ModuleNotFoundError: No module named 'django.core.urlresolvers'
```

```
Name: django-polymorphic
Version: 1.3
Summary: Seamless Polymorphic Inheritance for Django Models
Home-page: https://github.com/django-polymorphic/django-polymorphic
Author: Christopher Glass
Author-email: tribaal@gmail.com
License: BSD
```

https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax

tony:
Two additional warnings need handling:

```
Using Python version 2.7.1 from /Users/me/work/python/django-polymorphic/.tox/py27-django110/bin/python2.7
Using Django version 1.10.8 from /Users/me/work/python/django-polymorphic/.tox/py27-django110/lib/python2.7/site-packages/django
/Users/me/work/python/django-polymorphic/.tox/py27-django110/lib/python2.7/site-packages/django/db/models/base.py:349: RemovedInDjango20Warning: Managers from concrete parents will soon qualify as default manage
rs if they appear before any other managers in the MRO. As a result, 'base_objects' declared on 'polymorphic.PolymorphicModel' will no longer be the default manager for 'tests.MROBase2' in favor of 'objects' dec
lared on 'tests.MROBase1'. You can redeclare 'base_objects' on 'MROBase2' to keep things the way they are or you can switch to the new behavior right away by setting `Meta.manager_inheritance_from_future` to `Tr
ue`.
  if not opts.managers or cls._requires_legacy_default_manager():
/Users/me/work/python/django-polymorphic/.tox/py27-django110/lib/python2.7/site-packages/django/db/models/base.py:349: RemovedInDjango20Warning: Managers from concrete parents will soon qualify as default manage
[...]

PR #327, opened 2017-12-06
update MIDDLEWARE variable

tony:
This is compatible with Django 1.10+, and required for Django 2.0

> You’ll need to adapt old, custom middleware and switch from the MIDDLEWARE_CLASSES setting to the new MIDDLEWARE setting to take advantage of the improvements.

source: https://docs.djangoproject.com/en/1.11/releases/1.10/

related #325 #324


tony:
Closing, duplicate of  #318

Issue #204, opened 2016-03-29
Bulk create support


tng10:
Hi guys,

There's a ticket that already solved a problem for normal inheritance classes in django (https://code.djangoproject.com/ticket/24997), but in django_polymorphic it doesn't work.

Is there anyone facing the same problem?

Thanks in advance!

tng10:
Any idea?

vdboor:
I'm not sure why it fails, but this likely needs some support to fix the `polymorphic_ctype` field that we assign during `.save()`.

tng10:
Thanks @vdboor I will check this out!

MightySCollins:
Did anyone make any progress on this? It would be great to be able to bulk_create.

krzysztofgromadzki:
I started working on this. I will make a pull request soon.

AlanCoding:
What's the current status of this functionality? I see a PR was merged, but the documentation still states

https://django-polymorphic.readthedocs.io/en/stable/api/polymorphic.managers.html

> Multi-table models are not supported.

bckohan:
This is an upstream django issue.

PR #336, opened 2018-01-03
Bulk create support

krzysztofgromadzki:
Related issue: django-polymorphic/django-polymorphic#204

This implementation just invokes `pre_save_polymorphic` method from `PolymorphicModel` for each saving object just before django's `bulk_create` . Usually this method is used in `save` method implementation in `PolymorphicModel`. Particullary `pre_save_polymorphic` method checks that foreign key to `ContentType` is set. If not, sets it.
Other possible solution (I didn't test it) is `moving pre_save_polymorphic` call to `__init__` in `PolymorphicModel` and removed this call from save method. In this case overidded `bulk_create` won't exist

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -127,6 +127,12 @@ class PolymorphicQuerySet(QuerySet):
     as_manager.queryset_only = True
     as_manager = classmethod(as_manager)
 
+    def bulk_create(self, objs, batch_size=None):
+        objs = list(objs)
+        for obj in objs:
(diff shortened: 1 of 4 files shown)


vdboor:
Cool! I think this is quite a nice solution! Django's `bulk_create` also performs a `list(objs)` so there is no performance issue here. While moving the `pre_save_polymorphic` logic to `PolymorphicModel.__init__()` could be possible, you'll have to be careful with overriding models. This solution avoids such traps! :-)

PR #345, opened 2018-03-13
Update managers.rst

T4m:
The PolymorphicManager must be instanciated


vdboor:
Thanks, you're completely right!

PR #353, opened 2018-06-28
Fix prefetch_related behavior

mgrrx:
I stumbled upon https://github.com/django-polymorphic/django-polymorphic/issues/68 which causes some errors for us. As mentioned in the issue, the problem is cause by reusing the instances stored in base_result_objects_by_id. This PR tries to work around it by maintaining the lists differently which allows to make shallow copies of the real instances (see line 417). Feel free to test!

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -323,12 +323,12 @@ class PolymorphicQuerySet(QuerySet):
         Finally we re-sort the resulting objects into the correct order and
         return them as a list.
         """
-        ordered_id_list = []    # list of ids of result-objects in correct order
-        results = {}            # polymorphic dict of result-objects, keyed with their id (no order)
+        resultlist = []            # polymorphic list of result-objects
 
         # dict contains one entry per unique model type occurring in result,
         # in the format idlist_per_model[modelclass]=[list-of-object-ids]
         idlist_per_model = defaultdict(list)
+        indexlist_per_model = defaultdict(list)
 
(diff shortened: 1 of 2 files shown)


mgrrx:
*ping*

mgrrx:
I ported the test from #250 to the new test structure

mgrrx:
ping again ;)

vdboor:
@mgrrx Sorry for replying so late. This code looks really nice, but I worry about the usage of the indices. There is a change that a subclass table doesn't have a record for the base table. Would that break the indices?

mgrrx:
I don't really get the problem. If you send me an example I volunteer to look into it ;)

ctk3b:
Friendly ping! Would love to see a fix for this merged

imomaliev:
Bump

vdboor:
I've merged this! I also fixed an edge-case where the queryset retrieval would crash if a table was truncated. (see f769ed756800140d210abb2a4d4413a1a68daa05 and 4f89a879c0071879b4c823d0ff8e297cb1ec0077)

imomaliev:
Hi @vdboor, I would like to help, also we could transfer it under https://github.com/jazzband supervision

auvipy:
@vdboor I have monthly 40 hours paid open source time for python. Rather than transferring you can ad both of us for maintaining this.

auvipy:
my work relies heavily on this, please invite me as member of this org

PR #355, opened 2018-07-27
WIP - Support for django 2.1

auvipy:


auvipy:
@vdboor guidance wanted for fixing the deprecation warnings and django 2.1 test failures.

vdboor:
hi @auvipy thanks for looking into this issue. Unfortunately, having so many style issue fixes makes it nearly impossible to see what is changed specifically for Django 2.1. If you can, could you file a PR with only the changes required for Django 2.1?

I've applied autopep8 in the past, so when you let it ignore the line length the code would look pretty good. Transitioning from the max line length requires some manual intervention to keep it readable, and I would opt for using 100 or 120 chars instead of 80.

auvipy:
yes I can And I will

soon:
Seems like django 2.1 already added to travis: https://github.com/django-polymorphic/django-polymorphic/commit/2d1ad3f77363278af6b2b1e5e6a8a77b75a1b299

vdboor:
I'm closing this, as it seems to fix style errors only (we can use black for that later)

Issue #361, opened 2018-08-25
ManagerInheritanceWarning raised when default manager is a subclass of PolymorphicManager


godardth:
Hi,

I am using polymorphic-tree and get the ManagerInheritanceWarning when using a subclass of PolymorphicManager as my default manager in my PolymorphicMPTTModel subclasses.

Aware that this is more a polymorphic-tree issue but I think it can be fixed here in polymorphic by adding a check as below before raising the warning :

`if not isinstance(manager, PolymorphicManager) and not manager.__class__.__name__ == 'TreeManager':`

Any thoughts on that?

Thanks,
Theo

PR #362, opened 2018-08-25
Fix for issue #361

godardth:


vdboor:
This needs a better fix, since it allows to do `class TreeManager(..)` like django-mptt does and skip the checks here. I think we could inspect the `_queryset_class` instead.

Issue #202, opened 2016-03-14
get_real_instance_class throws TypeError


korkmaz:
There is a chance that `polymorphic_ctype_id` of an object can point a wrong non-proxy ContentType model. If that case happens polymorphic model's get_real_instance_class throws a TypeError instead of RuntimeError.

issubclass method is called with a None argument like `issubclass(model_instance, None)`

``` python
# Protect against bad imports (dumpdata without --natural) or other
# issues missing with the ContentType models.
if model is not None \
and not issubclass(model, self.__class__) \
and not issubclass(model, self.__class__._meta.proxy_for_model):
    raise RuntimeError("ContentType {0} for {1} #{2} does not point to a subclass!".format(
        self.polymorphic_ctype_id, model, self.pk,
    ))
    return model
```

Maybe above code should also check `self.__class__._meta.proxy_for_model` is a None or not before calling issubclass?

vdboor:
Hi thanks for reporting this! Could you write a test that replicates this problem? That would help a lot to write a proper solution.

amitnabarro:
I am experiencing similar problems. While i am able to reproduce in a particular system, i don't know how to re-create this in an encapsulated environment

amitnabarro:
I found out a descrepency in the polymorphic_ctype_id values in the base class. I still don't know what caused the polymorphic object to point to the wrong CotentType object, but at least it explains why the call fails

thenewguy:
I am hitting this too.  Have you solved the problem?

thenewguy:
Some how for me it appears that the base class is being saved.  Then an exception occurs during creating the child.  And the base class stays in the database but you can't delete it due to this error.  Not sure how this is happening

thenewguy:
or perhaps the child class is being deleted and the base class is not.  But some how the base class exists without a child references.  I am verifying this in the admin using the polymorphic filters

amitnabarro:
I did find out what my problem was for me, which was unrelated to django-polymorphic directly. I am implementing a multi-tenant system using Postgres schemas. As it turns out, you need to be really careful with content types when doing that. Since django-polymorphic relies on it (polymorphic_ctype_id in the base class) one has to make sure that if multiple schemas do not use the same content-type table, then clearing the content-type cache at every request is important. The only problem related to django-polymorphic was the cryptic error message

thenewguy:
Hrm... well I am only using one content type table.

[...]

PR #369, opened 2018-11-22
Raise `PolymorphicTypeInvalid` instead of cause `TypeError`

marksweb:
Enhancement for  #202 where a `TypeError` is caused by an invalid content type id.
This change allows that scenario to raise the `PolymorphicTypeInvalid` exception making the issue explicit.


vdboor:
Thanks!

PR #370, opened 2018-11-24
Restore compatibility with current versions of extra-views (issue #367).

### LINKED ISSUE #367: contrib/extra_views.py is not compatible with django-extra-views 0.11.0 or later
There are API changes in django-extra-views 0.11.0 or later which break compatibility with contrib/extra_views.py:

- 0.11.0 removed the ability to specify 'extra' in the class as in PolymorphicFormSetMixin.
- 0.12.0 renamed InlineFormSet to InlineFormSetFactory.

Would a PR that addressed these two issues be acceptable?

ShaheedHaque:
This resolves #367.

### DIFF
--- a/polymorphic/contrib/extra_views.py
+++ b/polymorphic/contrib/extra_views.py
@@ -25,7 +25,7 @@ class PolymorphicFormSetMixin(object):
     formset_class = BasePolymorphicModelFormSet
 
     #: Default 0 extra forms
-    extra = 0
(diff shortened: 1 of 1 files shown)


ShaheedHaque:
Hi, is there any chance of a release including this (and other?) PRs anytime soon?

vdboor:
Thanks for the patch! Sorry for not checking this project any sooner

PR #379, opened 2019-04-03
changes made to support djnago 2.2

auvipy:
work in progress, should django 1.11 support dropped?

PR #386, opened 2019-05-24
Support Django 2.2

mumumumu:
Following up on work started in https://github.com/django-polymorphic/django-polymorphic/pull/379 by @auvipy.

- add fix to `formsets.utils.add_media` to support Django 2.2
- add Django 2.2 to test matrix
- add missing middlewares to get tests working on Django 2+

The master branch tests are failing because Django has started removing support for python2, so `django.utils.six` is no longer available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,25 +3,6 @@ sudo: false
 language: python
 dist: xenial
 
-python: "3.6"
-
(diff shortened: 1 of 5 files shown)


mumumumu:
Yep, will do.

Issue #198, opened 2016-02-12
select_related and prefetch_related for inherited models


eriktelepovsky:
Regarding the documentation:

> select_related() works just as usual, but it can not (yet) be used to select relations in inherited models (like ModelA.objects.select_related('ModelC___fieldxy') )

Is there a chance you will implement this? I mean select_related and prefetch_related for the inherited models, It would be really helpful. Because I don't know how to optimize some requests now. Thank you.

vdboor:
Thanks for the report! You've identified a spot where polymorphic isn't present yet. I'm not sure this can be solved; the `.filter(ModelC___fieldxy=..)` already is somewhat of a hack. I'm not sure we can hook into the select-related logic but you're welcome to give it a try - granted that you'l add unit tests for it.

gregmuellegger:
I played around with select related in the child models, and figured I need to create a manager for the `base_objects` attribute on the child models, that performs a `select_related` for every queryset.

However the `select_related` config will get overriden by this line: https://github.com/django-polymorphic/django-polymorphic/blob/master/polymorphic/query.py#L351

I think that if this line didn't exist, we would have a good result that we can do `select_related` on child models.

I traced the history to 2010 and previous of this line. It seems to be there since the beginning. Could we do something like:

```
if not real_objects.query.select_related:
    real_objects.query.select_related = self.query.select_related
```

So that we don't override the select related config here?

tleguijt:
Are there any plans on supporting this and/or is any help needed?
Supporting select_related and prefetch_related would greatly improve the performance of some of our code that uses polymorphic models.

vdboor:
I like the assignment of the `select_related`. We'll have to make sure though that it doesn't trigger Django errors when that field doesn't exist in a particular subclass. A well tested pull request on this is welcome!

gregmuellegger:
Actually I think it *shall* trigger any Django errors if something is wrong. The `real_objects` queryset comes from the `base_objects` manager. And the `base_objects` is used specifically for retrieving objects of the subclass itself. So polymorphic shouldn't do any magic in this place.

pablolmedorado:
You would make me the happiest man in the world if there was a solution to this. I have many relationships in child models and the number of queries is too high.

Thank you for your work @vdboor!

hobarrera:
While working on #244, I uncovered something that might be useful, assume these models:

```python
class Parent(PolymorphicModel):
    pass

class ChildA(Parent):
[...]

PR #395, opened 2019-07-30
One-level select_related

OskarPersson:
This is an attempt at adding support for `select_related` and thus closing #198. So far it has no unit tests or documentation and has mostly been tested with [Graphene-Django](https://github.com/graphql-python/graphene-django) which is where I found my need for this. 

(Getting [graphene-django-optimizer](https://github.com/tfoxy/graphene-django-optimizer) to work with this is the next step!)

With my limited testing it seems to work at the first level (e.g. `Parent.objects.select_related('childa__related')`) but not any deeper than that (e.g. `Parent.objects.select_related('childa__related__childb')`)

This PR is not in a mergeable state and instead is currently mostly created for bringing the discussion further for implementing `select_related` with a more complete solution.

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -410,7 +410,21 @@ class PolymorphicQuerySet(QuerySet):
                 **{("%s__in" % pk_name): idlist}
             )
             # copy select related configuration to new qs
-            real_objects.query.select_related = self.query.select_related
+
+            if self.query.select_related is False:
+                real_objects.query.select_related = False
+            else:
+                concrete_model_name = real_concrete_class._meta.model_name
+                sub_model_names = set([m._meta.model_name for m in self.model.__subclasses__()])
+                sub_model_names.remove(concrete_model_name)
+
+                for sub_name in sub_model_names:
+                    self.query.select_related.pop(sub_name, None)
+
+                if concrete_model_name in self.query.select_related:
+                    real_objects.query.select_related = self.query.select_related[concrete_model_name]
+                else:
+                    real_objects.query.select_related = self.query.select_related
 
             # Copy deferred fields configuration to the new queryset
             deferred_loading_fields = []


ADR-007:
You can try something like that:
```python
from typing import List

from polymorphic.models import PolymorphicModel
from polymorphic.query import PolymorphicQuerySet


class SelectRelatedPolymorphicQuerySet(PolymorphicQuerySet):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.used_select_related_cache = False

    def use_select_related(self):
        qs = self._clone()
        qs.used_select_related_cache = True

        return qs

    def _clone(self, *args, **kwargs):
        qs = super()._clone(*args, **kwargs)
        qs.used_select_related_cache = self.used_select_related_cache

        return qs

    def _get_real_instances(self, base_result_objects: List[PolymorphicModel]):
        if self.used_select_related_cache:
            return [item.get_real_instance() for item in base_result_objects]

        return super()._get_real_instances(base_result_objects)
```
```python
from polymorphic.managers import PolymorphicManager
from polymorphic.models import PolymorphicModel

from vas.utils.models.polymorhic.querysets import SelectRelatedPolymorphicQuerySet


class SelectRelatedPolymorphicModel(PolymorphicModel):
    class Meta:
        abstract = True

    objects = PolymorphicManager.from_queryset(SelectRelatedPolymorphicQuerySet)()

    def get_real_instance(self):
        real_model = self.get_real_instance_class()

        if real_model == self.__class__:
            return self

        # allow to use cache:
        # return getattr(self, real_model.__name__.lower())
        return self._state.fields_cache[real_model.__name__.lower()]  # todo
```

bckohan:
Thanks for this start @OskarPersson!

I am the new leader of this project. It will take me a bit to work through the backlog of issues but this feature is also a desire of mine so I plan to revisit this as time allows. If you want to take another look that would be great too!

bckohan:
I am closing this PR in favor of either #545 or #531

PR #399, opened 2019-08-19
Fix for projects which might have a different base_url or script_name.

joshuamaxwell:
The request.path isn't the correct path to match the url patterns with is it? Shouldn't this use reqeust.path_info so that if someone's application is hosted as a subdirectory you will still be matching the correct portion of the path in the django urls?

With this code in its present state any admin integration will bring my application down.


vdboor:
thanks!

PR #403, opened 2019-09-09
Change `docs/quickstart.rst` version support to match `README.rst`.

loganknecht:


loganknecht:
Hey there,

I know it's a bit a small PR, but I was reading the documentation here:
https://django-polymorphic.readthedocs.io/en/stable/quickstart.html#

I got worried about the version support of the project not working with my current django project, and then ended up at the `README.rst` here that said otherwise.

More than happy to throw this away if it's not wanted, but just trying to help keep the documentation in sync is all :)

loganknecht:
I'm not sure what the reason is for failure on this. I've only changed some text.

chrisglass:
I think travis was not in a good shape at the time, and the maintainership has been... lackluster (sorry).

Merging this now.

PR #413, opened 2019-11-06
Use TypedChoiceField for polymorphic_ctype field

### LINKED ISSUE #412: BasePolymorphicModelFormSet should use TypedChoiceField in add_fields()
In the [BasePolymorphicModelFormSet add_fields method](https://github.com/django-polymorphic/django-polymorphic/blob/master/polymorphic/formsets/models.py#L231), a ChoiceField is added to the form. The single choice in this field references the `polymorphic_ctype` of the model that this form is for. 

When this form is POSTed back to django, the value in this field is parsed as a string, where the initial option was an int. This breaks the [has_changed](https://docs.djangoproject.com/en/2.2/ref/forms/fields/#has-changed) method in the Field base class, since it compares the initial value (an int) to the newly returned value (a string). 

The fix is trivial - use a TypedChoiceField instead, and to pass it `coerce=int`. I can create a PR for this later tonight.

bo0tzz:
Fixes #412

### DIFF
--- a/polymorphic/formsets/models.py
+++ b/polymorphic/formsets/models.py
@@ -228,8 +228,8 @@ class BasePolymorphicModelFormSet(BaseModelFormSet):
             form._meta.model, for_concrete_model=False
         )
         choices = [(ct.pk, ct)]  # Single choice, existing forms can't change the value.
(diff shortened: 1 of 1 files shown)


vdboor:
good catch! Sorry for not noticing this PR before

PR #418, opened 2019-12-03
Pass `widgets` field to `modelform_factory`

### LINKED ISSUE #417: PolymorphicFormSetChild does not pass widgets argument to modelform_factory
The PolymorphicFormSetChild class [takes a `widgets` argument](https://github.com/django-polymorphic/django-polymorphic/blob/master/polymorphic/formsets/models.py#L38), but does not [pass it when calling `modelform_factory`](https://github.com/django-polymorphic/django-polymorphic/blob/master/polymorphic/formsets/models.py#L84). This makes it impossible (?) to specify custom widgets for these forms.

bo0tzz:
Fixes #417


vdboor:
thanks! sorry for not noticing this PR

PR #435, opened 2020-05-15
Non polymorphic parent

### LINKED ISSUE #9: Placing non polymorphic model after a polymorphic model can cause AttributeError no base_objects
Sometimes I'll get an error like this:

AttributeError  type object XXX has no attribute 'base_objects'

The class XXXX is always one that inherits from models.Model, and generally later in my models file than a polymorphic class.   Reordering the class file to put the normal models further up fixes it.

Here's an example trace, with 'Region' which extends model.Model causing the error:
https://gist.github.com/2966844

Model of Region:

``` python
class Region(models.Model):
    name = CharField(max_length=256)

    def __unicode__(self):
        return self.name
```

foarsitter:
Hello Folks, 

Here is my proposal to fix #9. 

When we have a PolymorphicModel with a non-polymorphic parent, deleting the parent is missing `_base_objects`. This pull-requests ads a fallback to `objects` when `_base_objects` is None. 

Cheers!

### DIFF
--- a/polymorphic/models.py
+++ b/polymorphic/models.py
@@ -205,7 +205,8 @@ class PolymorphicModel(with_metaclass(PolymorphicModelBase, models.Model)):
 
         def create_accessor_function_for_model(model, accessor_name):
             def accessor_function(self):
-                attr = model._base_objects.get(pk=self.pk)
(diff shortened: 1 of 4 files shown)


vdboor:
thanks! nice and subtle fix.

Issue #437, opened 2020-05-20
Foreign key resolves to parent class when using abstract models


mcovalt:
Foreign key lookups don't upcast to their actual type when a parent polymorphic model inherits from an [abstract base class](https://docs.djangoproject.com/en/3.0/topics/db/models/#abstract-base-classes). This is inconsistent with the behavior described [here](https://django-polymorphic.readthedocs.io/en/stable/advanced.html#manytomanyfield-foreignkey-onetoonefield).

Example:
```python
from django.db import models
from polymorphic.models import PolymorphicModel

class CommonInfo(models.Model):
    class Meta:
        abstract = True

class Parent(CommonInfo, PolymorphicModel):
    fk = models.ForeignKey("self", on_delete=models.CASCADE, null=True)

class Child(Parent):
     field = models.CharField(max_length=10)


related_obj = Child.objects.create(field="related child")
obj = Child.objects.create(fk=related_obj, field="child")
obj.refresh_from_db()
type(obj.related_obj) == Child    # False
type(obj.related_obj) == Parent   # True
```

mehmetakyuz:
Had the same issue. Moving the abstract class right of the PolymorphicMode in the class definition also resolves it. 
So PolymorphicModel should be kept as first for the method resolution order.

trybik:
Note: this wasn't an issue for `django-polymorphic<2`, so if you're migrating you need to make sure to adjust inheritance order like @mehmetakyuz noted above.

bckohan:
Whats happening here is that Django uses the manager at the attribute Meta.base_manager_name or if none is specified it creates a simple Manager() object to access related objects. When walking the inheritance chain to determine if a base_manager_name is specified it only looks for the first parent's base manager name so MRO order isn't exactly respected in that if no base_manager_name is found it doesnt keep looking at lower priority parents.

A possible fix here is to set base_manager_name to "objects" in the `__new__` of PolymorphicModel if it is currently None.

PR #438, opened 2020-05-20
Test that FK lookup resolves child when using an AbstractModel

mcovalt:
Test for #437


bckohan:
I had trouble commiting to this branch even though it seems permission was given. I took these commits, fixed the issue and am replacing this PR with #679

Issue #439, opened 2020-05-23
`tox -e docs` fails


berhoel:
`tox -e docs` fails with:

```
Warning, treated as error:
html_static_path entry '_static' does not exist
```

A `mkdir docs/_static` fixes the command.

chrisglass:
The fix was merged.

PR #440, opened 2020-05-23
Fix `tox -e docs` for out or the box checkout.

berhoel:
This fixes issue #439 (`tox -e docs` fails) for me.


chrisglass:
Ah, a simple yet effective fix indeed.

Thanks! merging.

Issue #424, opened 2019-12-20
Django 3.0.1 support


daneb255:
In Django 3.0.1 there is a DeprecationWarning with the ugettext_lazy function.

RemovedInDjango40Warning: django.utils.translation.ugettext_lazy() is deprecated in favor of django.utils.translation.gettext_lazy().

changed to:
from django.utils.translation import gettext_lazy as _

I would be very happy about a fix :) I can not create a new branch :/

Greetings

holgerco:
In Django 3.0 
`django.utils.six` deprecated
please fix it

daneb255:
Push :)

vdboor:
> RemovedInDjango40Warning: django.utils.translation.ugettext_lazy() is deprecated in favor of django.utils.translation.gettext_lazy().

That is correct. Replacing it breaks Python 2 compatibility, so it would mean we release a new Python-3-only release.

> In Django 3.0 django.utils.six deprecated please fix it

We're not using django.utils.six in polymorphic :-)

The current release already runs on Django 3. So, while this issue needs to be addressed in the future, it's a low-prio right now.

AndreyMZ:
1.  > it would mean we release a new Python-3-only release.

	It is not necessary. The alternative is to use conditional imports.
	
2.  Python 2 compatibility is not needed anymore since there are not any supported Django versions  supporting Python 2. (See <https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django>.)

3.  > The current release already runs on Django 3.

	Actually, no, it doesn't. See #443.
	
So, I think the priority of this bug should be raised greatly.

chrisglass:
I think this should be fixed on master after the merging of #452

As far as I understand it, this means the next release of polymorphic will be python3 only.

yakky:
@chrisglass master is already python 3 only :)
I opened #457

chrisglass:
@yakky yes, that is exactly what I am saying :) master is python3 only, therefore the next release also will.

PR #441, opened 2020-05-23
Got rid of `RemovedInDjango40Warning` warnings mentioned in #424

berhoel:
Silenced

```
RemovedInDjango40Warning: django.utils.translation.ugettext_lazy() is deprecated in favor of django.utils.translation.gettext_lazy().
```

warnings from django 3 mentioned in #424 . Should still work for Django2.

tox fails with the same errors as a clean clone from the repository.

### DIFF
--- a/example/orders/models.py
+++ b/example/orders/models.py
@@ -1,7 +1,11 @@
+import django
 from django.db import models
 from django.utils.dates import MONTHS_3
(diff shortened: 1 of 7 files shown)


Djailla:
This is now ok in master. You can close this PR ;)

PR #445, opened 2020-06-10
Performance improvement on parentadmin

safaariman:
Instead of querying the database for every content type, we can simply get them in one query.


chrisglass:
Seems to help indeed! thanks!

PR #451, opened 2020-08-04
Corrected import location

OmenApps:
I noticed this import was pointing to the wrong location when it was called out after I upgraded to django 3.1. Looking at query.py, it imports the FieldDoesNotExist exception from django.db.models, rather than from django.core.exceptions, where FieldDoesNotExist is actually defined.

In the past, django.db.models had a reference to FieldDoesNotExist, so importing from that location was not an issue, but at some point in the past that reference was removed, so the import fails.


OmenApps:
Closing. #452 is more complete.

PR #459, opened 2020-08-18
Make test_defer_fields use get_deferred_fields()

chrisglass:
Use the provided get_deferred_fields() method instead of relying on the
repr() for the objects to determine whether a field is deferred or not.

This is an attempt at fixing the flaky travis runs (since I assume repr() relies on dict ordering somehow, although I have not checked).


chrisglass:
This makes travis happy it seems (https://travis-ci.org/github/django-polymorphic/django-polymorphic/builds/718965887)

Issue #455, opened 2020-08-14
Failures on 3.2


yakky:
Apparently this change https://github.com/django/django/commit/9c9a3fe1180fc92fbd4c3302dbe0b3e083bf0381 in django master broke custom version of PolymorphicQuerySet._filter_or_exclude

Upstream change is weird, and we need to investigate the rationale before adapting our code

AdamDonna:
I've taken a quick look at this one and it looks like the api of `_filter_or_exclude` has change to take args, kwargs and negate as their respective types rather than retaining the old api.

This passes all tests for django master and above but isn't backwards compatable because the api of _filter_or_exclude has changed.

```
    def _filter_or_exclude(self, negate, args, kwargs):
        # We override this internal Django functon as it is used for all filter member functions.
        q_objects = translate_polymorphic_filter_definitions_in_args(
            queryset_model=self.model, args=args, using=self.db
        )
        # filter_field='data'
        additional_args = translate_polymorphic_filter_definitions_in_kwargs(
            queryset_model=self.model, kwargs=kwargs, using=self.db
        )
        args = list(q_objects) + additional_args
        return super(PolymorphicQuerySet, self)._filter_or_exclude(
            negate=negate, args=args, kwargs=kwargs
        )
```

I'm new to opensource so there a standard method of proxying we can use to be backwards compatible with django less than 3.2?

AdamDonna:
^^ Actually I have an idea. I'll have a PR up soon and we can discuss the relative merits of the approach on that and go from there 👍

PR #461, opened 2020-08-19
Fixes issue 455

AdamDonna:
Fixes https://github.com/django-polymorphic/django-polymorphic/issues/455
Raised on django master in https://github.com/django/django/commit/9c9a3fe1180fc92fbd4c3302dbe0b3e083bf0381

I'd like specific feedback on this approach of handling django versions and any documentation that will be necessary for maintenance in the future.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -13,6 +13,7 @@ Contributors
 * Abel Daniel
 * Adam Chainz
 * Adam Wentz
+* Adam Donaghy
 * Andrew Ingram (contributed setup.py)
 * Al Johri
 * Alex Alvarez
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -157,7 +157,7 @@ class PolymorphicQuerySet(QuerySet):
         # Implementation in _translate_polymorphic_filter_defnition."""
         return self.filter(not_instance_of=args)
 
-    def _filter_or_exclude(self, negate, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


yakky (commented):
Handling of the different django API looks perfectly reasonable  to me

I would just indent the `if get_django_version() ...` in line with the other class methods, but it's a matter of taste

AdamDonna:
I'm not sure we can indent that line it since it's operating on the class, I'm happy for you to push it up if you know how to make it work inline with the class definition 👍

yakky:
> I'm not sure we can indent that line it since it's operating on the class, I'm happy for you to push it up if you know how to make it work inline with the class definition +1

Sorry, the diff misled me

Something like

```python
if get_django_version() >= "3.2":
   def _filter_or_exclude(....):
      ....
else:
   def _filter_or_exclude(....):
      ....
```
in the class body would work

See this example https://github.com/divio/django-filer/pull/1201/files#diff-71f149c65e62bb3b8fdcd50b4b00a845R105

yakky:
But I would wait for @chrisglass input to check if he agrees on the approach

AdamDonna:
Yeah cool, I never knew that was a thing. I much prefer that way so related methods are together and it's not a shock for future devs that there's a 2 definitions of `_filter_or_exclude` to handle
I've got that passing locally, it's ready to go depending on if @chrisglass agrees with the approach

AdamDonna:
@chrisglass How do you feel about this solution?

Issue #463, opened 2020-09-09
PyPI sdist lacks tests


jayvdb:
The sdist only contains a few tests files, missing the actual tests

```
...
./django-polymorphic-3.0.0/polymorphic/templatetags/polymorphic_formset_tags.py
./django-polymorphic-3.0.0/polymorphic/tests
./django-polymorphic-3.0.0/polymorphic/tests/migrations
./django-polymorphic-3.0.0/polymorphic/tests/migrations/0001_initial.py
./django-polymorphic-3.0.0/polymorphic/tests/migrations/__init__.py
./django-polymorphic-3.0.0/polymorphic/utils.py
...
```

vdboor:
the tests are explicitly excluded. I've extended this rule to avoid cluttering the install

jayvdb:
Normal users install the wheel, which is not the sdist and wont include these tests.  The sdist is able to be used by distros who can set up a proper test rig. e.g. https://repology.org/project/python:django-polymorphic/versions . fwiw I am one of those package maintainers.  Anyway, thanks for working through the backlog.

PR #464, opened 2020-09-09
MANIFEST.in: Update and add tests

jayvdb:
Closes https://github.com/django-polymorphic/django-polymorphic/issues/463


vdboor:
the tests can't be run without a proper test setup. This won't work and would clutter the site-packages on install.

PR #489, opened 2021-05-18
Update Python Versions in Documentation

rickstc:
# Overview
As I was preparing to create pr #488 I was reviewing the contribution guidelines and noticed that the python versions were pretty out of date compared to the information in tox.

This just updates the documentation.


AdamDonna (approved):
Yeah looks like it's a good update to reflect reality as specified in https://github.com/django-polymorphic/django-polymorphic/blob/master/tox.ini#L2

vdboor:
thanks!

PR #496, opened 2021-08-29
PolymorphicParentModelAdmin.add_type_view edge case

matijakolaric:
In some cases (mine :P) self.get_child_type_choices(request, "add") returns [], so this case must be handled.


vdboor:
good catch!

PR #499, opened 2021-10-10
Django 4.0a compatibility

fraimondo:
Use force_str instead of force_text (deprecated in Django 3.2, removed in Django 4.0)


vdboor:
thanks!

Issue #503, opened 2021-11-29
Missing setuptools runtime dependency


markgras:
The `django-polymorphic` package's top-level `__init__.py` file imports `pkg_resources`, which is part of the `setuptools` package. Attempting to use `django-polymorphic` in a Python environment that does not have `setuptools` installed leads to an `ImportError`. To fix this, `django-polymorphic`'s metadata should properly state that `setuptools` is a runtime dependency of the package.

jonprindiville:
Hey, @markgras ... just a shot in the dark here, I don't know if I bumped into the same thing as you or not, but I **did** experience a similar symptom: django-polymorphic's import of `pkg_resources` threw an `ImportError`.

It was in CI (with a work project) and it was pretty odd to me because I wasn't changing anything related to django-polymorphic at the time, I was moving some other dependency from Poetry's list of "dev" deps to the "main" list of deps.

Anyhow, in **my** case what feels like ended up being the problem was that I had crossed up the versions of Poetry I was using locally in CI: basically, there were two places where we configured the Poetry version and I only updated one, then goofy things happened when I made dependency changes 💥 

That issue when away for me when I corrected my _using two different versions of Poetry_ goof.

Don't know if you experienced a similar thing, but though I'd share.

markgras:
@jonprindiville You ran into this GitHub issue, as `pkg_resources` is part of `setuptools` ([source code](https://github.com/pypa/setuptools/tree/main/pkg_resources)). In your problem configuring Poetry you ended up creating an environment that didn't have `setuptools` installed (or perhaps installed in the wrong place). Seeing this issue pop up in the wild is rare because `setuptools` is already present in most environments for old `python setup.py install` installation backward-compatibility. However, as more and more packages become PEP 517/518-compliant, we may see this backward-compatibility dropped, in which case this issue will become more apparent.

jonprindiville:
> you ended up creating an environment that didn't have `setuptools` installed

Nope. I just thought I'd share a possible trigger for this issue in case you were stuck.

> However, as more and more packages become PEP 517/518-compliant, we may see this backward-compatibility dropped, in which case this issue will become more apparent.

Ok, cool, that would have been useful information to include in the description of the issue.

Carry on. 👋

PR #504, opened 2021-11-29
Add setuptools dependency in setup.cfg

markgras:
Addresses issue #503


sliverc:
As an update with `pkg_resources` being deprecated I have developed a PR #541 where there is no need for setuptools to be used. This would make adding `setuptools` as a dependency obsolete as well.

markgras:
I like that solution more! I will close this PR in favor of yours.

PR #539, opened 2023-04-08
Fix multi-fields in single line

morsik:
As [Django documentation](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets) states, you may use:

```python
{
    "fields": [("first_name", "last_name"), "address", "city", "state"],
}

```

to specify that `first_name` and `last_name` should be displayed in single line.

Django Polymorphic though wasn't able to work with this syntax properly as it tried to remove whole tuple instead of individual elements from child fields.

This PR fixes this, and now multi-field lines are rendered correctly.

### DIFF
--- a/polymorphic/admin/childadmin.py
+++ b/polymorphic/admin/childadmin.py
@@ -227,7 +227,15 @@ class PolymorphicChildModelAdmin(admin.ModelAdmin):
 
         # Find which fields are not part of the common fields.
         for fieldset in self.get_base_fieldsets(request, obj):
-            for field in fieldset[1]["fields"]:
+            # multiple elements in single line
+            if isinstance(field, tuple):
+                for line_field in field:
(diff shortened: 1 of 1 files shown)


morsik:
@jleclanche have you checked this PR? I totally forgot about this, but after reading it right now again, it seems I did big mistake in it!

In this PR I removed `for field in fieldset[1]["fields"]:` in this PR, but I think it should stayed as it was, and all my code and else statement with original code should be just inside for loop...

jleclanche:
@j-antunes could you submit the corrected version you merged recently?

morsik:
@jleclanche fixed, see PR above ^

j-antunes:
@morsik - Awesome thank you!

PR #541, opened 2023-04-25
Removed usage of deprecated pkg_resources

### LINKED ISSUE #503: Missing setuptools runtime dependency

sliverc:
Fixes #503 

`pkg_resources` is deprecated and should not be used anymore. We encountered this in our CI see https://github.com/django-json-api/django-rest-framework-json-api/pull/1146

The version is now defined in `polymorphic/__init__.py` and `setup.py` extracts it from there. This way, there is also no need for setuptools to be installed when using django polymorphic.

### DIFF
--- a/polymorphic/__init__.py
+++ b/polymorphic/__init__.py
@@ -6,9 +6,7 @@ This code and affiliated files are (C) by Bert Constantin and individual contrib
 Please see LICENSE and AUTHORS for more information.
 """
 
-import pkg_resources
+VERSION = "3.1.0"
 
-try:
-    __version__ = pkg_resources.require("django-polymorphic")[0].version
-except pkg_resources.DistributionNotFound:
-    __version__ = None  # for RTD among others
+# version synonym for backwards compatibility
+__version__ = VERSION
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = django-polymorphic
-version = 3.1.0
+version = attr: polymorphic.VERSION
 description = Seamless polymorphic inheritance for Django models
 long_description = file:README.rst
 author = Bert Constantin


markgras:
A few things:
1. Did you give the `attr:` directive a try in `setup.cfg`? I think that that would be a cleaner drop-in replacement, would be less prone to error than a custom regex, and would keep all project metadata in `setup.cfg` instead of splitting it up between that and `setup.py`. [Reference](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html).
2. [PEP 396](https://peps.python.org/pep-0396/) has been rejected, so perhaps the best solution here is actually to make a breaking change and remove the `__version__` attribute in `__init__.py` altogether.

sliverc:
Thanks @markgras for the hinter. I wasn't aware of attr directive. I have updated it accordingly. Furthermore, I added a version synonym for backwards compatibility, but if the maintainers decide this is not needed I am happy to remove it as well.

markgras:
Happy to help! The module attribute being accessed can have any name. In your new commit you could assign the version string to `__version__` directly and then access that attribute in `setup.cfg` if you'd like. I believe that `setuptools`' example purposely doesn't use a dunder variable so as to not unintentionally imply that the rejected PEP 396 should be followed.

sliverc:
Just as an update. Python 3.12 removed pkg_resources, so it is not possible to use django-polymorphic anymore with Python 3.12. This PR fixes this as well. Great if it could get merged.

sgabb (approved):

martin-thoma:
> polymorphic/__init__.py:9: DeprecationWarning: pkg_resources is deprecated as an API

The last release is from 2021: https://pypi.org/project/django-polymorphic/#history , but in this repository are many updates.

Is djanog-polymorphic dead (on pypi)?

IgorArnaut:
It's baffling that the update won't still be on PyPI.

martin-thoma:
@bckohan Do you know when a PyPI release will be done? What are the blockers?

bckohan:
> @bckohan Do you know when a PyPI release will be done? What are the blockers?

I can have one up as soon as [this goes through.](https://github.com/jazzband/help/issues/388)

I just need to review the diffs from the last release.

Issue #542, opened 2023-05-09
TypeError: cannot pickle '_thread.lock' object


rrauenza:
I'll start with I don't know WHY the Q object has a _thread.lock object in it ... but ...

I think this is caused by this code, and the deepcopy of the Q object:

`    translate_polymorphic_Q_object(queryset_model, copy.deepcopy(q), using=using) for q in args
`

I wonder if the code should be using q.clone() (undocumented?) or even better, q.all() (which does the same as q.clone()), which supposedly makes a copy of the query?

I noticed this seems similar to this issue:

https://github.com/django-polymorphic/django-polymorphic/issues/524

that originates at the same line of code.

rrauenza:
I have a proposed PR that has a custom deepcopy to avoid these pickle issues.

rrauenza:
full stack trace, for posterity.  It isn't clear how this _thread object appears in the deepcopy...

```
  File ".../.venv/lib/python3.11/site-packages/polymorphic/query_translate.py", line 95, in translate_polymorphic_filter_definitions_in_args
    return [
           ^
  File ".../.venv/lib/python3.11/site-packages/polymorphic/query_translate.py", line 96, in <listcomp>
    translate_polymorphic_Q_object(queryset_model, copy.deepcopy(q), using=using) for q in args
                                                   ^^^^^^^^^^^^^^^^
  File "/opt/python-3.11.3/lib/python3.11/copy.py", line 153, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File ".../.venv/lib/python3.11/site-packages/django/utils/tree.py", line 53, in __deepcopy__
    obj.children = copy.deepcopy(self.children, memodict)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python-3.11.3/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/opt/python-3.11.3/lib/python3.11/copy.py", line 206, in _deepcopy_list
    append(deepcopy(a, memo))
           ^^^^^^^^^^^^^^^^^
  File "/opt/python-3.11.3/lib/python3.11/copy.py", line 146, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/opt/python-3.11.3/lib/python3.11/copy.py", line 211, in _deepcopy_tuple
[...]

PR #543, opened 2023-05-09
Replace a deepcopy of the Q object with Q.all()

rrauenza:


rrauenza:
Still trying to run unit tests locally, hoping marking this PR as ready would trigger some on project side.

rrauenza:
See https://github.com/django-polymorphic/django-polymorphic/issues/542

TomHaii:
Can also confirm that is indeed fixing the issue in deepcopying Q objects. (reproduced on Python 3.7)

rrauenza:
I think if the existing unit tests work, that is sufficient, since it shows the change doesn't break from the current method of copying.

We could assign a non picklable object attr to one of the children to prevent future regression?

AdamDonna:
+1 to adding a non picklable object attr to prevent future regressions or even an explicit scenario that should break in 3.7 without this fix

rrauenza:
@AdamDonna It's been a while -- I came back to try to add a test.

The test is successful, and it fails if I swap deepcopy back in.  I'm not happy with the test data,  I could not figure out how to trigger the code path into django's trees.py except through this method.

Can we accept this PR so I can eventually remove my patch from my project?

rrauenza:
Q() derives from tree.Node, which adds copy in Django 4.2+, and does this (from django source):

```
   def __copy__(self):
        obj = self.create(connector=self.connector, negated=self.negated)
        obj.children = self.children  # Don't [:] as .__init__() via .create() does.
        return obj

    copy = __copy__
```

Let's jump into the debugger and see what happens.


```
polymorphic/tests/test_admin.py .
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home.local/rich/src/django-polymorphic/polymorphic/query_translate.py(99)_deepcopy_q_object()
-> obj = q.copy()
(Pdb) pp q
<Q: (OR: ('one2one__in', [<Model2D: id 1, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField)>]))>
(Pdb) 
(Pdb) pp q.children
[('one2one__in',
  [<Model2D: id 1, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField)>])]
```

Step past q.copy() and we get:

```
(Pdb) pp obj
<Q: (OR: ('one2one__in', [<Model2D: id 1, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField)>]))>
(Pdb) pp obj.children
[('one2one__in',
  [<Model2D: id 1, field1 (CharField), field2 (CharField), field3 (CharField), field4 (CharField)>])]
(Pdb) 
```

Are these objects actually deep copied in django's copy()?

No.  both the tuple and the Model are just references to the original.  So my implementation for Django < 4.2 doesn't do anything different, IMHO.

```
(Pdb) id(obj.children[0][1][0])
140227184738432
(Pdb) id(q.children[0][1][0])
140227184738432
(Pdb) id(obj.children[0])
140364539161792
(Pdb) id(q.children[0])
140364539161792
```

rrauenza:
There is one oddity, though, that I've highlighted in my latest push.  Django's copy seems to change the case of some of the Q() parameters.  See the comment in the code for more info.

bckohan:
Test failures are a result of changes introduced by #572

Issue #390, opened 2019-06-27
Proxy Model Support - Creating base class and querying for proxy model does not return results


Eluzive:
Hello,

First off thanks for creating this package!

I am running into an issue now where we needed to create two different admin pages and creating proxy models seems like a good idea to use the same model have this separation or grouping. However, when registering the proxy models to their own separate ModelAdmins, the get_queryset() returns an empty set and results are not shown.

I've posted on [SO](https://stackoverflow.com/questions/56796864/empty-queryset-when-using-a-polymorphic-model-from-django-polymorphic-and-queryi)

When I tried creating the object using the regular Django Model class, I am able to query the proxy model and get the created object.

```
from django.db import models

class Person(models.Model):
    first_name = models.CharField(max_length=30)
    last_name = models.CharField(max_length=30)

class MyPerson(Person):
    class Meta:
        proxy = True

    def do_something(self):
        # ...
        pass
```

```
>>> p = Person.objects.create(first_name="foobar")
>>> p
<Person: Person object (4)>
>>> MyPerson.objects.all()
<QuerySet [<MyPerson: MyPerson object (4)>]>

```
However, when I use the PolymorphicModel and query for proxy model, the queryset is empty.
```

class Person(PolymorphicModel):
    first_name = models.CharField(max_length=30)
    last_name = models.CharField(max_length=30)
```

```
>>> p = Person.objects.create(first_name="foobar")
>>> MyPerson.objects.all()
<PolymorphicQuerySet []>
```

Shouldn't we expect the Queryset to contain the Person object we created?

thejpanganiban:
We're also encountering this issue. :(

Any updates on this?

ParikhKadam:
Any updates??

emyller:
This is [by design](https://github.com/django-polymorphic/django-polymorphic/blob/5b9acd184868e634240a728580e7732f459e99cd/polymorphic/managers.py#L31-L32). For now you can work around this by overriding the manager's `get_queryset`:

```python
class MyModelManager(PolymorphicManager):
    def get_queryset(self):
        qs = self.queryset_class(self.model, using=self._db, hints=self._hints)
        return qs


class MyModel(models.Usuario):
    objects = MyModelManager()

    class Meta:
        proxy = True
```

pannkotsky:
@emyller can you explain that design? in which case `qs.instance_of(self.model)` can return anything for proxy model when fetching objects from DB? it seems impossible

PR #544, opened 2023-06-26
Add option to model meta class to prevent proxy models being polymorphic

### LINKED ISSUE #376: Proxy model for base class needs manager definition
Hello,
I have a class with a custom manager defined by:
`objects = PolymorphicManager.from_queryset(PointOfInterestQuerySet)()`

The proxy model:
```
class PointOfInterestGlobalSummary(PointOfInterest):
    class Meta:
        proxy = True
        verbose_name = _('Dashboard - Global')
        verbose_name_plural = _('Dashboard - Global')
```
Running some queries:
```
>>> PointOfInterest.objects.all().count()
139534
>>> PointOfInterestGlobalSummary.objects.all().count()
0
```

The solution was to define the proxy model with a normal manager:
`objects = models.Manager()`
The queries:
```
>>> PointOfInterest.objects.all().count()
139534
>>> PointOfInterestGlobalSummary.objects.all().count()
139534
```

Is this normal behavior? If so, maybe we should add some documentation?

Cheers

pgammans:
Add support for 'proxy' polymorphic models to behave in the same manor as Django models where no supplementary table is created, and that queries on the proxy class are the same as those on it base,

Fixes #376 #390

### DIFF
--- a/polymorphic/base.py
+++ b/polymorphic/base.py
@@ -57,6 +57,12 @@ class PolymorphicModelBase(ModelBase):
     def __new__(self, model_name, bases, attrs, **kwargs):
         # print; print '###', model_name, '- bases:', bases
 
+        polymorphic__proxy = None
+        if "Meta" in attrs:
+            if hasattr(attrs["Meta"], "polymorphic__proxy"):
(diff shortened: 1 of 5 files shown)


bckohan:
Superseded by #676

PR #531, opened 2022-10-13
Support select_related and prefetch_related for inherited models

k4anubhav:
add `select_polymorphic_related` and `prefetch_polymorphic_related` methods for select and prefetch on inherited models

Triggers Django errors when that field doesn't exist in a particular subclass

Example works exactly same as django select_related and prefetch_related
```python
queryset = Project.objects.select_polymorphic_related(
    ArtProject, 'artist', 'canvas__painter'
).select_polymorphic_related(
    ResearchProject, 'supervisor',
)


queryset = Project.objects.prefetch_polymorphic_related(
    ArtProject, 'artist', Prefetch('canvas', queryset=Project.objects.annotate(size=F('width') * F('height')))
).prefetch_polymorphic_related(
    ResearchProject, 'authors',
)
```

Issue #436, opened 2020-05-20
hacks for getting around the `select_related` issue from parent on inherited models?


drivelous:
So I've read this issue extensively and tried some things out which don't work -- https://github.com/django-polymorphic/django-polymorphic/issues/198

So I'm trying to see if I can come to a middle ground I have some models defined like this

```
class Message(PolymorphicModel):
    body = models.TextField
    # fields

class SMSMessage(Message)
    from_phone = models.CharField
    to_phone = models.ForeignKey(FooPhoneNumberModel)

class EmailMessage(Message):
    from_email = models.CharField
    to_email = models.ForeignKey(FooEmailModel)

class InstantMessage(Message):
    from_username = models.CharField
    to_username = models.ForeignKey(FooUser)

```

Since it's not supported, I don't need the free query that comes from `select_related` e.g.

```
communications = Communication.objects.select_related('smsmessage', 'emailmessage')
```

I'm mostly concerned with the n + 1 queries when serializing these. Right now every instance adds a query to the count

`<CommunicationQuerySet: [<SMSMessage>, <EmailMessage>, <SMSMessage>]>`

I was trying to mess around and see if I could run some type of prefetch? Something like:

```
communications = Communication.objects.prefetch_related(
    Prefetch('smsmessage', queryset=SMSMessage.objects.filter(user=user), to_attr='through__smsmessage'),
    Prefetch('emailmessage', queryset=EmailMessage.objects.filter(user=user), to_attr='through__emailmessage'),
)
```

so that the query can execute the same amount of queries whether it's 20 messages or 1000 messages.

Any help is appreciated including explanations of the internals. I've been trying to do things like call `select_related('smsmessage')` and `select_related('communication_ptr__smsmessage') and the like and my mind is mush right now

Issue #359, opened 2018-08-09
Can't prefetch base instance


lerela:
Hi,

I am struggling to understand how to make my queries more efficient when the parent object must be accessed (use case that led me here is described below).

Here is a basic scenario:

    from django.db import models
    from polymorphic.models import PolymorphicModel

    class Base(PolymorphicModel):
        field1 = models.CharField(max_length=8)

    class Child1(Base):
        field2 = models.CharField(max_length=8)

All I want to do is to load both Child1 _and_ the base instance so that I can access both Child1 & its Base counterpart.

I create my object with:

    child1 = Child1(field1="a", field2="b")

And I fetch it with:
    
   child1 = Child1.objects.filter(pk=1)[0]

This triggers the following SQL query:

     SELECT "test_polymorphic_base"."id", "test_polymorphic_base"."polymorphic_ctype_id", "test_polymorphic_base"."field1", "test_polymorphic_child1"."base_ptr_id", "test_polymorphic_child1"."field2" FROM "test_polymorphic_child1" INNER JOIN "test_polymorphic_base" ON ("test_polymorphic_child1"."base_ptr_id" = "test_polymorphic_base"."id") WHERE "test_polymorphic_child1"."base_ptr_id" = 1  LIMIT 1; args=(1,)

Note that both field1 and field2 were successfully SELECTed, as well as any field allowing to directly instanciate the Base instance.
Calling `child1.field1` and `child1.field2` do not trigger additional query, as expected.

However, calling `child1.base_ptr` does:

    SELECT "test_polymorphic_base"."id", "test_polymorphic_base"."polymorphic_ctype_id", "test_polymorphic_base"."field1" FROM "test_polymorphic_base" WHERE "test_polymorphic_base"."id" = 1; args=(1,)

Note that all those fields were _already_ fetched with previous query. So I figure I'd use `.select_related`, but the result is the same:

    child1 = Child1.objects.select_related('base_ptr').filter(pk=1)[0] # child1.base_ptr triggers an extra query

I also tried `.prefetch_related`:

    child1 = Child1.objects.prefetch_related('base_ptr').filter(pk=1)[0]

Which fails immediately with:

    ValueError: 'base_ptr' does not resolve to an item that supports prefetching - this is an invalid parameter to prefetch_related().

This seems to be related to the fact that `Child1.base_ptr` does not have a `.get_prefetch_queryset()` method.

[...]

Issue #244, opened 2016-09-28
select_related() from a plain model with a ForeignKey to a polymorphic model


Bartvds:
If you have a regular model that has a ForeignKey to a polymorphic model, and then use select_related() from the regular model''s manager/queryset to select the related polymorphic instances we only get these downcasted as the base class of the polymorphic model.

Illustration:

````python
class ParentModel(PolymorphicModel):
    pass

class ChildModel(ParentModel):
    # some fields etc
    pass   

class PlainModel(models.Model):
    relation = models.ForeignKey(ParentModel)
    
    
# if we let it do 1+n
for obj in PlainModel.objects.all():
    # these obj.relation values will have their proper sub type
    print(type(obj.relation))
        
# if we select_related()
for obj in PlainModel.objects.select_related('relation'):
    # these obj.relation values will all be ParentModel's
    print(type(obj.relation))
````

My quess is the PlainModel's manager is not aware it has to apply the polymorphic magic? Is this expected behaviour? I see the note on select_related() here http://django-polymorphic.readthedocs.io/en/stable/advanced.html but I think it doesn't apply.

**Comment by lorinkoz** [comment]
Same issue here.

**Comment by hobarrera** [comment]
The linked note seems to refer to #198. This sounds like a separate (although similar-sounding) issue.

**Comment by hobarrera** [comment]
~I've been looking into this the last couple of hours. Looks like it's a lot harder than it seems.~

~`select_related` fetches from multiple tables in a single query, and this is constructed [here](https://github.com/django/django/blob/b9f7dce84b7ab5e198129030eae6c1a4aec83d24/django/db/models/sql/compiler.py#L177). This is then constructed into instances using `from_db` [here](https://github.com/django/django/blob/b9f7dce84b7ab5e198129030eae6c1a4aec83d24/django/db/models/query.py#L1616).~

A hackier approach may work, since is that it's possible to do a weird `select_related` call that fetches subchildren, so it's also possible to produce a hacky workaround:

```python
qs = PlainModel.objects.select_related('relation__childmodel')

real_instances = []
for obj in qs:
    try:
        obj.relation = obj.relation.childmodel
    except AttributeError:
        pass
    # You can also try other subclasses, just add them to the select_related() call above.
```

Note that this raises `AttributeError`: it doesn't try to query the DB, is just fails an attribute lookup (which is **a lot** cheaper!).

[...]

PR #545, opened 2023-06-26
Related polymorphic query support inc select_related and prefetch_related

### LINKED ISSUE #498: Prefetch's to_attr field is not respected for polymorphic models
If I `prefetch_related` with a `to_attr` field, the results of the prefetch do not seem to get saved to an attribute on the query results if the prefetch is for a polymorphic model.

So, using an example from the Django docs, if one does:
```
voted_choices = Choice.objects.filter(votes__gt=0)
prefetch = Prefetch('choice_set', queryset=voted_choices, to_attr='voted_choices')
Question.objects.prefetch_related(prefetch).get().voted_choices
```
...the final line will result in an error if Choice is a polymorphic model. Using `not_polymorphic()` on the queryset doesn't seem to help.

pgammans:
This pull request add support for select related and general related queries.

**Summary**
To take advantage of this it requires that the plain Django model either adds a custom query set manager and uses that for queries where it wants the related object to be transmogrify into the final polymorphic type of
that the default manager's queryset uses/ has PolymorphicQuerySetMixin or PolymorphicRelatedQuerySet as a base class

ie
`
class PlainModel(models.Model):
    objects = models.Manager.from_queryset(PolymorphicRelatedQuerySet)()
`

The default Django select_related can be used including which means this is simple to use from the Django admin.

**Performance**
When using select_related we archive fetch of the data in a single query including all information required to transmogrify the related object. This is an improvement from n+1 (main line) or models + 1 (with #531)

**Notes**
* This make a small change to the accessor function on the model allowing the base object to be fetched form the cached
* This commit has to copy segments of Django internal for it RelatedPopulator and ModelIterable as there are no hooks. 

It has minimal support for prefetch_related cannot fetch attributes not on all child models or via class names this may fix #498

Fixes: #198 #436 #359 #244

### DIFF
--- a/polymorphic/models.py
+++ b/polymorphic/models.py
@@ -198,9 +198,18 @@ class PolymorphicModel(with_metaclass(PolymorphicModelBase, models.Model)):
         self.__class__.polymorphic_super_sub_accessors_replaced = True
 
         def create_accessor_function_for_model(model, accessor_name):
+            NOT_PROVIDED = object()
+
             def accessor_function(self):
-                objects = getattr(model, "_base_objects", model.objects)
-                attr = objects.get(pk=self.pk)
+                attr = NOT_PROVIDED
+                try:
+                    attr = self._state.fields_cache[accessor_name]
+                    pass
+                except KeyError:
+                    pass
+                if attr is NOT_PROVIDED:
+                    objects = getattr(model, "_base_objects", model.objects)
+                    attr = objects.get(pk=self.pk)
                 return attr
 
             return accessor_function
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -2,19 +2,25 @@
 QuerySet for PolymorphicModel
 """
 import copy
+import functools
+import operator
 from collections import defaultdict
 
 from django import get_version as get_django_version
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import FieldDoesNotExist
 from django.db.models import FilteredRelation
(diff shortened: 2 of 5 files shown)


PetrDlouhy:
@pgammans Thanks, this seems to be very useful improvement.
Could you please rebase the code so that the tests would run against the current master?

Also I thing part of this PR should be change of documentation to reflect this, e.g.: https://django-polymorphic.readthedocs.io/en/stable/advanced.html?highlight=select_related#restrictions-caveats

Morfild:
Are you going to merge it one day?

bvedad:
What is the status of this MR?

bckohan:
> pgammans

This is on the top of my list to review. Other reviews are also welcome!

bckohan:
@pgammans Would you be so kind as to check the Allow edits by maintainers button? Otherwise I will need to create a new PR from this one.

pgammans:
@bvedad I am not able to do this as far as i know due to (https://github.com/orgs/community/discussions/5634)

pgammans:
I have a version of this re-based onto master , other than fixing merge errors is also addresses an issues caused by PR #620.

See #758 but in summary this PR changes the meaning of the polymorphic_primary_key_name, it is no long the name of the field in the local model that points to the prime / base polymorphic object pk and has become just a  Model._meta.pk which isn't the always the field on an inherited model

I have work around the issues #758 so this pass all test but but without a fix for it it make the code a little more complex as we need to track the field the map from the model used in on a relation and the final polymorphic object.

PR #550, opened 2023-10-25
Multi-database support in inheritance accessors.

ghost:
This was an issue that kept popping up in a work project that utilizes django-polymorphic. Specifically, the issue was that django-polymorphic didn't respect multiple databases for deletion, causing it to repeatedly error out if the object didn't exist in the `default` database, but did exist in the `target_db`.

I've verified that this change passes the tests, as outlined in the community contribution guidelines.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,7 @@ Changelog
 v4.3.0 (202X-XX-XX)
 -------------------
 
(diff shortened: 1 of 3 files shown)


j-antunes:
Hi @Spikiii - The project has recently moved under a new team. If possible could you rebase this PR against `master`? 

Also, could you provide a unit test for you changes? Thanks!

PR #562, opened 2023-12-12
Use pytest for testing & modernize assertions

akx:
This switches to pytest + pytest-cov + pytest-django for testing, and uses https://github.com/akx/codemod-unittest-to-pytest-asserts to codemod all `self.assertSomethingSomething` style assertions to plain assertions, which Py.test can rewrite for a better DX.


j-antunes (approved):

j-antunes:
Thanks for your awesome work on this! I'll leave it up this PR up for a little longer just in case someone wants to also review it. Later in the day let's start merging your open PRs

PR #563, opened 2023-12-12
Drop support for non-supported versions of Django and Python

### LINKED ISSUE #556: Make Django 3.2 (LTS) lowest supported version
Django 3.2 is the lowest current version supported by Django https://endoflife.date/django
We should drop support up to 3.2 and add support for 4.2

akx:
Resolves #556

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,21 +22,16 @@ jobs:
           name: dist
           path: dist/
   Test:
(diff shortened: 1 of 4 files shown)


j-antunes (approved):
Thanks for your work on this PR!

jgadelange:
Am I right that didn't make it into the README and documentation?

akx:
@jgadelange You're right! Sorry about that.

PR #566, opened 2023-12-20
Replace string interpolation and .format calls with f-strings

akx:
Guided by `ruff --select=UP` + [`flynt`](https://github.com/ikamensh/flynt/), hand-checked afterwards.


j-antunes:
Thanks for you work in this PR!

PR #572, opened 2024-01-31
Check if child node is Q instance for subquery expressions

CelestialGuru:
Fixes bug when using `Exists` subqueries which are or-ed together:

```python
SomeModel.objects.filter(
   Exists(RelatedModelA.objects.filter(some_model=OuterRef("pk"))
   | Exists(RelatedModelB.objects.filter(some_model=OuterRef("pk"))
)
```

This is done usually for performance reasons, but it is a valid queryset. `Exist` instances ored together like this creates a `Q` instance, but the children are not themselves `Q` instances. They are `Exists` instances.

### DIFF
--- a/polymorphic/query_translate.py
+++ b/polymorphic/query_translate.py
@@ -69,7 +69,7 @@ def translate_polymorphic_Q_object(queryset_model, potential_q_object, using=DEF
                 )
                 if new_expr:
                     node.children[i] = new_expr
-            else:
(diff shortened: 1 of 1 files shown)


j-antunes:
Thanks for opening a PR. Could you add the following:
* A unit test to cover your changes
* In your PR you change `else` to an `elif`, could you add another `else` so that in case none of the conditions are met, it will raise an exception or return something else.
Thank you

PR #582, opened 2024-02-29
Auto-discover child models and inlines

### LINKED ISSUE #576: Add `PolymorphicInlineModelAdmin.get_child_inlines()`
The same as `PolymorphicParentModelAdmin.get_child_models()` allows to dynamically get the child models with `base_model.__subclasses__()`, I would like to have `PolymorphicInlineModelAdmin.get_child_inlines()` method that I can overwrite to get the inlines subclasses, and don't need to set them explicitly. This would allows me to have a cleaner code, and also prevent some circular references.

piranna:
Allow by default to inspect for Model and Inline subclasses to populate the admin forms. Also raise a proper exception in case there are no child models defined.

Fixes #576.

### DIFF
--- a/polymorphic/admin/inlines.py
+++ b/polymorphic/admin/inlines.py
@@ -19,7 +19,7 @@ from polymorphic.formsets import (
 )
 from polymorphic.formsets.utils import add_media
 
-from .helpers import PolymorphicInlineSupportMixin
+from .helpers import PolymorphicInlineSupportMixin, get_leaf_subclasses
 
 
 class PolymorphicInlineModelAdmin(InlineModelAdmin):
@@ -52,7 +52,7 @@ class PolymorphicInlineModelAdmin(InlineModelAdmin):
 
     #: Inlines for all model sub types that can be displayed in this inline.
     #: Each row is a :class:`PolymorphicInlineModelAdmin.Child`
-    child_inlines = ()
+    child_inlines = None
 
     def __init__(self, parent_model, admin_site):
         super().__init__(parent_model, admin_site)
@@ -76,12 +76,42 @@ class PolymorphicInlineModelAdmin(InlineModelAdmin):
         for child_inline in self.child_inline_instances:
             self._child_inlines_lookup[child_inline.model] = child_inline
 
+    def get_child_inlines(self):
+        """
+        Return the derived inline classes which this admin should handle.
+        This should return a list of tuples, exactly like :attr:`child_inlines` is.
+
+        The inline classes can be retrieved as ``base_inline.__subclasses__()``,
+        a setting in a config file, or a query of a plugin registration system at your option
+        """
+        if self.child_inlines is not None:
+            return self.child_inlines
+
+        child_inlines = get_leaf_subclasses(PolymorphicInlineModelAdmin.Child)
+        child_inlines = tuple(
+            inline
+            for inline in child_inlines
+            if (
+                inline.model is not None and
+                issubclass(inline.model, self.model)
+            )
+        )
+
+        if child_inlines:
+            return child_inlines
+
+        raise ImproperlyConfigured(
+            "No child inlines found for '{self.model.__name__}', please "
+            "define the 'child_inlines' attribute or overwrite the "
+            "'get_child_inlines()' method."
+        )
+
     def get_child_inline_instances(self):
         """
         :rtype List[PolymorphicInlineModelAdmin.Child]
         """
         instances = []
-        for ChildInlineType in self.child_inlines:
+        for ChildInlineType in self.get_child_inlines():
             instances.append(ChildInlineType(parent_inline=self))
         return instances
 
--- a/polymorphic/admin/parentadmin.py
+++ b/polymorphic/admin/parentadmin.py
@@ -123,9 +123,14 @@ class PolymorphicParentModelAdmin(admin.ModelAdmin):
         Return a list of polymorphic types for which the user has the permission to perform the given action.
         """
         self._lazy_setup()
+
+        child_models = self.get_child_models()
+        if not child_models:
+            raise ImproperlyConfigured("No child models are available.")
+
         choices = []
         content_types = ContentType.objects.get_for_models(
-            *self.get_child_models(), for_concrete_models=False
+            *child_models, for_concrete_models=False
         )
 
         for model, ct in content_types.items():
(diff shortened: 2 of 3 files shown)


j-antunes:
Could you also add a unit test?

piranna:
> Could you also add a unit test?

Where could I add them? `polymorphic/tests` folder?

polymorphic/admin/inlines.py
```diff
@@ -53,7 +53,8 @@ class PolymorphicInlineModelAdmin(InlineModelAdmin):
 
     #: Inlines for all model sub types that can be displayed in this inline.
     #: Each row is a :class:`PolymorphicInlineModelAdmin.Child`
-    child_inlines = ()
+    child_inlines = None
+    exclude_childs = None
```
akx, line 57:
```suggestion
    exclude_children = None
```

This also needs a doc comment to go along with it.

polymorphic/admin/inlines.py
```diff
@@ -53,7 +53,8 @@ class PolymorphicInlineModelAdmin(InlineModelAdmin):
 
     #: Inlines for all model sub types that can be displayed in this inline.
     #: Each row is a :class:`PolymorphicInlineModelAdmin.Child`
```
akx, line 55: The semantics of `None` will need to be documented here.

I also don't think this should be automatic by default – it could mean that upgrading Polymorphic in a project suddenly exposes things in the admin that one might not want to expose.

polymorphic/admin/parentadmin.py
```diff
@@ -50,6 +51,7 @@ class PolymorphicParentModelAdmin(admin.ModelAdmin):
 
     #: The child models that should be displayed
     child_models = None
+    exclude_childs = None
```
akx, line 54:
```suggestion
    exclude_children = None
```
(and docs again, please?)

polymorphic/admin/inlines.py
piranna: Why do you propose the change of `childs` for `children`? `childs` is used everywhere else in the source code...

polymorphic/admin/inlines.py
piranna: Why don't you think they should be exposed? The current alternative is that code crash on them.

polymorphic/admin/parentadmin.py
piranna: Same as before, why the change?

polymorphic/admin/inlines.py
akx: Where can you find `childs`? I (or `git grep`) certainly can't:

```
~/b/django-polymorphic (master) $ git grep children | wc -l
      70
~/b/django-polymorphic (master) $ git grep childs | wc -l
       0
```

polymorphic/admin/inlines.py
akx: Sorry, I'm not following. That the code crashes on... what exactly?

piranna: They Crash if there are no child classes defined.

piranna:
Sorry for keep this open so much time, I have been focused on other parts of our project, but I think we are ready to get this PR merged. What else is needed to do?

polymorphic/admin/inlines.py
akx: I still think `children` would be the correct spelling btw. :)

piranna: Changed, I think I got confused with the usage of `child` (single) everywhere. You are right, `childs` is not used in any other part of the code, and also `children` is the gramatically correct alternative, thank you for drawing attention on this :-)

piranna: Also added the documentation.

piranna:
Fixed concerns and linting, only missing one are tests, but I don't see how to run them. Anyway, we have already this PR running on production on our systems.

bckohan:
Thank you for your submission! I have reviewed the code and appreciate what you want to do here. I am going to accept that we should provide a ```get_child_inlines()``` hook in inline admin forms analogous to the ```get_child_models()``` hook in the parent admin. 

I am more hesitant to include the default class introspection behavior you have added. I agree that it is very convenient, but I think there are corner cases and potential pitfalls to making it the default behavior. I can see us including it in the future, but if we do so we'll need a robust set of tests. To that end I'm going to start a discussion and reference this PR so we don't lose track of it.

I didn't have permission to push to this branch so #681 replaces this PR - I'm going to add the ```get_child_inlines()``` hook which you should be able to override in extensions to the admin classes to add in this dynamic discovery behavior to your projects.

piranna:
Please reopen, PR #681 doesn't replaces this PR, and although a discussion can reference this, being closed doesn't show It as active anymore.

bckohan:
> Please reopen, PR #681 doesn't replaces this PR, and although a discussion can reference this, being closed doesn't show It as active anymore.

I appreciate your effort and I have incorporated it into #681 . I am closing this PR for three reasons:

1. #681  still provides the hook that you can override in subclasses of ```PolymorphicInlineModelAdmin``` - which you can then use in your project with the implementation of ```get_child_inlines``` provided here. I think this resolves the main issue highlighted in  #576 - which is the lack of a hook to more easily implement whatever dynamic behavior you are looking for without a monkey patch. 
2. No tests were added.
3. The implementation of ```get_child_inlines()``` here would find every ```PolymorphicInlineModelAdmin.Child``` class that has a model type that inherits from the parent ```PolymorphicInlineModelAdmin``` model class. This would break in projects that have multiple ```PolymorphicInlineModelAdmin.Child``` implementations for the same model type - which can happen when you want different admin behavior for different foreign keys pointing to the same model type in different models.

I think a safer implementation that uses nested class encapsulation might look like:

```python
from inspect import isclass

class Inline(StackedPolymorphicInline):
    model = InlineModelA

    def get_child_inlines(self):  # <-- put this implementation in a mixin
        return [
            child
            for child in self.__class__.__dict__.values()
            if isclass(child) and issubclass(child, StackedPolymorphicInline.Child)
        ]

    class InlineModelAChild(StackedPolymorphicInline.Child):
        model = InlineModelA

    class InlineModelBChild(StackedPolymorphicInline.Child):
        model = InlineModelB
        autocomplete_fields = ["plain_a"]
```

For further discussion please hop over to #682

PR #614, opened 2024-08-19
Fixed custom __init__ check in transmogrify

bdabrowski:
Fixes https://github.com/jazzband/django-polymorphic/issues/615


bckohan (approved):
The custom ``__init__`` transmogrify code was dead because it erroneously looked at the object dict to locate an ``__init__`` implementation instead of the class ``__dict__``.

I think there is still an edge case here where values set differently by the base class and subclass's ``__init__`` may be overridden after the subclass instantiation - but its hard to solve for that and this code is certainly more correct.

Thank you!

Issue #53, opened 2013-08-16
OneToOneField cannot be primary key


macropin:
This example

```
class Account(PolymorphicModel):
    user = models.OneToOneField(User, primary_key=True)
```

fails with `Caught AttributeError while rendering: type object 'Account' has no attribute 'polymorphic_primary_key_name'`

2c47db8fcc284a92d2c9769ba503603fbea92660 introduced a new way of determining pk's which specifically excludes OneToOneField fields:  `if f.primary_key and type(f)!=models.OneToOneField:`

This issue was reported downstream 2 years ago at https://github.com/bconstantin/django_polymorphic/issues/17.

Why exclude OneToOneField's? The tests continue to pass after removing this restriction.

DylanLukes:
I'm also having the exact same issue. If I understood the workings of Rest Framework well enough I'd submit a pull request but...

burrito-dev:
same issue

NeielJair:
same issue

dilawerh:
same issue

Onur1991:
Any updates on this issue?

piranna:
Any update on this? I think have got the same issue, and I'm surprised this has been open for 10 years, specially since there's a fix... How can we move this forward?

bckohan:
Fixed by #620

PR #620, opened 2024-12-08
Remove checking of OneToOneField type in determining primary key name.

sezabart:
Tripping over the same issue that was documented [here](https://github.com/jazzband/django-polymorphic/issues/53) back in 2013 and without a solution, I decided to get rid of the error-causing clause and live dangerously.

In my experience this works, but there might be edge cases (@vdboor mentions one) that this is supposed to protect against, which it won't now. But at least I can use my OneToOneField's

### DIFF
--- a/polymorphic/base.py
+++ b/polymorphic/base.py
@@ -73,7 +73,7 @@ class PolymorphicModelBase(ModelBase):
         # determine the name of the primary key field and store it into the class variable
         # polymorphic_primary_key_name (it is needed by query.py)
         for f in new_class._meta.fields:
-            if f.primary_key and type(f) != models.OneToOneField:
+            if f.primary_key:
                 new_class.polymorphic_primary_key_name = f.name
                 break
 


bckohan:
Needs tests

bckohan:
The name of the primary key field is needed by the polymorphic queries and is determined in ``__new__`` during polymorphic model class creation. The method of determination was incorrect - it was looping through the fields and picking the first field that is not a OneToOneField and has primary_key=True set. Django creates a OneToOneField for each parent class in multi-table inheritance and all of them, are marked as primary_key=True - these are logical primary keys that the base class machinery resolves into a single pk that is stored in _meta.pk. The fix here is to simply refer to _meta.pk.

This fixes #53 

All of the tests are passing and I do not expect this will trigger any untested regressions - but if anyone is hitting regressions with this fix you could revert to the old behavior by setting (and please also submit a bug report):

```python
class YourModel(PolymorphicParent):
    ...

YourModel.polymorphic_primary_key_name = "what it was before upgrade to 4.3"
```

Issue #616, opened 2024-09-16
Bug: Using Aggregations in Django >= 5.1.1 fails


suspectpart:
After updating to Django 5.1.1, I ran into errors using Aggregations:

```py
  File "/home/me/PycharmProjects/my-project/src/accounting/models.py", line 1440, in period
    return tuple(self.aggregate(Min("period_start"), Max("period_end")).values())
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/me/PycharmProjects/my-project/.venv/lib/python3.12/site-packages/polymorphic/query.py", line 316, in aggregate
    self._process_aggregate_args(args, kwargs)
  File "/home/me/PycharmProjects/my-project/.venv/lib/python3.12/site-packages/polymorphic/query.py", line 303, in _process_aggregate_args
    test___lookup(a)
  File "/home/me/PycharmProjects/my-project/.venv/lib/python3.12/site-packages/polymorphic/query.py", line 298, in test___lookup
    test___lookup(source_expression)
  File "/home/me/PycharmProjects/my-project/.venv/lib/python3.12/site-packages/polymorphic/query.py", line 300, in test___lookup
    assert "___" not in a.name, ___lookup_assert_msg
                        ^^^^^^
``` 

The line that breaks is:

https://github.com/jazzband/django-polymorphic/blob/cfd49b26d580d99b00dcd43a02409ce439a2c78f/polymorphic/query.py#L279

In Django 5.1.1 a change was made that consistently includes the filter arg of an expression in any case, even if it is empty - so if no filter is used, an additional `None` will be include in the list returned by `get_source_expressions()`. See here:

https://github.com/django/django/commit/42b567ab4c5bfb1bbd3e629b1079271c5ae44ea0#diff-c7d87b6e6b973bf87e5685442a6c1da9531c24bcaa86fe158c40ea0276986f32R53

It seems like we need an additional `None` check here, just like we have in `patch_lookup`:

https://github.com/jazzband/django-polymorphic/blob/cfd49b26d580d99b00dcd43a02409ce439a2c78f/polymorphic/query.py#L255

(This is also a quick workaround: If you run into this issue, transforming all aggregate expressions passed into `aggregate()` / `annotate()` into keyword args solves the problem).

Should I submit a quick PR to do that? Or is supporting Django 5 a broader issue?

aprams:
[...]

PR #621, opened 2025-02-05
PolymorphicQuerySet: Prevent NoneType error in test___lookup

hll1682003:
## Description:
This PR ensures that test___lookup only processes non-None source_expression values to prevent potential NoneType errors. The change adds a simple condition to check if source_expression is None before passing it to test___lookup.

## Changes:

Added a None check for source_expression before calling test___lookup.
Why this is needed:

- Prevents potential NoneType errors when evaluating query expressions.
- Ensures test___lookup is only invoked with valid values.
- Our unit tests start failing after migrating from Django 5.0.10 to 5.1.5 and start showing "AttributeError: 'NoneType' object has no attribute 'name'"
- Inspired by [this fix](https://github.com/jazzband/django-polymorphic/commit/88922d7d5e918f1338eeb9baeecb17ed98ed1261), which addressed a similar issue in django-polymorphic.

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -277,7 +277,8 @@ class PolymorphicQuerySet(QuerySet):
                 tree_node_test___lookup(self.model, a)
             elif hasattr(a, "get_source_expressions"):
                 for source_expression in a.get_source_expressions():
-                    test___lookup(source_expression)
+                    if source_expression is not None:
+                        test___lookup(source_expression)
             else:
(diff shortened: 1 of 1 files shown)


bckohan:
Fixes #616

bckohan (approved):
This fix is correct. The alternative would be to do the none check in the else statement below, but I can find no other places where test__lookup is called on a potentially None value in the code.

hll1682003:
Thanks @bckohan for your review!

PR #622, opened 2025-02-06
Use CSS variables for colors

joaoseckler:
More recent versions of Django admin use CSS variables to control the theming color, allowing a user to chose their color scheme independently from their preferred one. Previously, django-polymorphic supported different color schemes only via the `prefers-color-scheme` media query. This commit implements usage of CSS variables for colors, ensuring consistent compatibility with the rest of the admin interface.


bckohan (approved):
Looks good, thankyou!

PR #629, opened 2025-03-16
Django 5.1+ compatibility for PolymorphicQuerySet._process_aggregate_args()

LeGast00n:
Since Django v5.1, django.db.models.Aggregate.get_source_expressions() can have None as item in the returned list, causing the final assert to fail.


khink:
@LeGast00n Your MR looks very nice. i was running into this error myself on Django 5.2. Could you maybe fix the failing ruff check, so it can be merged?

@bckohan IS anything else needed to get this merged?

bckohan:
@khink @LeGast00n a test would be very helpful!

bckohan:
superseeded by #638

PR #632, opened 2025-04-20
Check for None on real_model in get_real_instance

nlsfnr:
Fixes a bug where `PolymorphicModel`'s `get_real_instance()` raises an exception when `get_real_instance_class()` returns None.

### DIFF
--- a/polymorphic/models.py
+++ b/polymorphic/models.py
@@ -157,8 +157,9 @@ class PolymorphicModel(models.Model, metaclass=PolymorphicModelBase):
         retrieve objects, then the complete object with it's real class/type
         and all fields may be retrieved with this method.
 
-        If the model of the object's actual type does not exist (e.g. it was
-        removed but its ContentType still exists), this method returns self.
+        If the model of the object's actual type does not exist (i.e. its
+        ContentType is stale), this method raises a
+        :class:`~polymorphic.models.PolymorphicTypeInvalid` exception.
 
         .. note::
             Each method call executes one db query (if necessary).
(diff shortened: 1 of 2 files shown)


bckohan:
@nlsfnr how did this situation arise?

nlsfnr:
Slightly weird scenario -- I use Docker containers for local development with a persitent volume for the DB. When switching between branches, where one branch has a new subclass of a polymorphic model that the other branch does not, the ContentType for said subclass still exists in the DB's volume but does not correspond to a Django model in the code base. That means that `get_real_instance_class` returns None and `get_real_instance` therefore crashes.

It's more or less equivalent to removing a Django model from the code base without adding the appropriate migration.

bckohan:
My only hesitation is that this might hide when something else is very wrong, like having unapplied migrations or other edge cases. I'm new to this project and still not the officially blessed maintainer so I'll revisit this when a decision is made on that.

thanks!

nlsfnr:
Yes good point.

I was mostly matching what was best for my use case, but looking at the implementation of `get_real_instance_class`, it may be more consistent to raise a `PolymorphicTypeInvalid` (or a new exception, e.g. `PolymorphicTypeUnknown`) exception instead.

If downstream applications prefer returning self, they could simply overwrite `get_real_instance`, which is what I am currently doing in my project.

Happy to change the PR accordingly.

bckohan:
PolymorphicTypeInvalid certainly seems appropriate.

bckohan (approved):

PR #634, opened 2025-04-20
Fix use of is and is not vs. ==

nlsfnr:
CI currently crashes with ruff complaining about the use of == to compare types. This breaks CI on unrelated PRs.


bckohan (approved):
This is correct, thank you!

PR #638, opened 2025-05-01
Django 5.1+ compatibility for PolymorphicQuerySet._process_aggregate_args()

khink:
@bckohan I tried to add a test for the change in https://github.com/jazzband/django-polymorphic/pull/629.
But i'm looking for some help, as unfortunately, the added code isn't hit by this test.

I'm running this with `tox -e py312-django52`. 
In order to run test just this test,  i changed `tox.ini` as follows:

```
[testenv]
setenv =
        ...
	PYTEST_ADDOPTS = -k test_polymorphic__aggregate_empty_queryset -s
 ```

The code in our project has these models (simplified):
```
class Article(PolymorphicModel):
    sales_points = models.IntegerField(verbose_name="Verkooppunten", null=True)  # type: ignore[var-annotated]

class AnalysisPackage(Article):
    name = models.CharField(max_length=100)

class ComponentListHeader(Article):
    name = models.CharField(max_length=100)

class OrderLine(TimeStampedModel, VenHLineMixin, models.Model):
    articles = models.ManyToManyField(  # type: ignore[var-annotated]
        Article, related_name="orderline", verbose_name="Artikelen"
    )

    def calculate_total_sales_points(self) -> int:
        article_points = self.articles.aggregate(Sum("sales_points"))[
            "sales_points__sum"
        ]
```

The error is triggered in a test in our project when we call `self.calculate_total_sales_points()`, more specific: In the `self.articles.aggregate(Sum("sales_points"))` part. In this case, there are no Articles / AnalysisPackages / ComponentListHeaders in the test database.

I tried copying these models in the test (`ModelArticle`, `ModelPackage`, `ModelComponent`, `ModelOrderLine`) but although the test passes, it does so without hitting the added line. In an interactive shell, i see that `test___lookup` is hit twice in the test, the values of `a` are `Sum(F(sales_points))` and  `F(sales_points)`.

Would you have any pointers?

### DIFF
--- a/polymorphic/query.py
+++ b/polymorphic/query.py
@@ -260,6 +260,9 @@ class PolymorphicQuerySet(QuerySet):
         def test___lookup(a):
             """*args might be complex expressions too in django 1.8 so
             the testing for a '___' is rather complex on this one"""
+            if a is None:
+                return  # since Django v5.1, get_source_expressions() return can have None in its items
+
             if isinstance(a, Q):
 
                 def tree_node_test___lookup(my_model, node):
--- a/polymorphic/tests/migrations/0001_initial.py
+++ b/polymorphic/tests/migrations/0001_initial.py
@@ -2064,4 +2064,86 @@ class Migration(migrations.Migration):
             },
             bases=("auth.group", models.Model),
         ),
+        migrations.CreateModel(
+            name="ModelArticle",
+            fields=[
+                (
+                    "id",
+                    models.AutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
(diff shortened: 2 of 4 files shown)


bckohan:
@khink Turns out this bug was already fixed by #621 which prevents Nones returned by get_source_expressions from ever being passed to test__lookup.

This does add a missing test however so I'm going to remove the redundant fix but keep the tests! thankyou!

This fix has not been deployed to pypi yet, but we'll get releases out hopefully no later than 2 weeks from now when [this is resolved](https://github.com/jazzband/help/issues/388).

bckohan:
superseded by #639

PR #648, opened 2025-08-20
Modernize build to use Hatchling

akx:
This gets rid of legacy `setup.py` and `setup.cfg` in favor of a modern PEP517 build using [Hatchling](https://hatch.pypa.io/latest/) as the build backend.

Hatchling was chosen as a backend because it has good defaults and it's easy to hook into – in our case, we need to `compilemessages` before bundling a wheel.

Running zipinfo on the .whl artifact generated by the GHA CI run shows the .mo files are there:

```
$ zipinfo dist/django_polymorphic-4.1.1-py3-none-any.whl | grep 'mo$'
-rw-r--r--  2.0 unx      337 b- defN 20-Feb-02 00:00 polymorphic/locale/en/LC_MESSAGES/django.mo
-rw-r--r--  2.0 unx      456 b- defN 20-Feb-02 00:00 polymorphic/locale/es/LC_MESSAGES/django.mo
-rw-r--r--  2.0 unx      492 b- defN 20-Feb-02 00:00 polymorphic/locale/fr/LC_MESSAGES/django.mo
```


bckohan (approved):

bckohan:
Hi @akx I'm going to take what you did here and add a switch over to uv to it. No need to resolve conflicts here - I'll supersede this with another PR

bckohan:
Hi @akx I'm going to take what you did here and add a switch over to uv to it. No need to resolve conflicts here - I'll supersede this with another PR

PR #649, opened 2025-08-20
Update pre-commit tools, fix ruff configuration so it doesn't warn

akx:
Ruff would complain about the legacy configuration on every run.

No more!


akx:
CI is broken in this PR on 3.8 due to a change in dj-database-url (see https://github.com/jazzband/dj-database-url/pull/271), but Python 3.8 is EOL anyway (see #650).

PR #676, opened 2025-12-04
Fix proxy model support.

### LINKED ISSUE #376: Proxy model for base class needs manager definition

bckohan:
Fixes #376 #390

I need to rebase and didn't have permissions to push to the original branch for #544, so this replaces that PR.

When polymorphic support was added to django-polymorphic circa 2013-2014 the design decision was to treat proxy models as concrete models wrt to polymorphism and to restrict the managers to those instances. This works because for_concrete_model=False is passed when the the ContentTypes are tagged onto the models - so if a row is created from the proxy model the proxy model's ContentType will be used to tag it. This is how you get situations described by #376:

```python
class MyModel(PolymorphicModel):
    ...

class MyProxy(MyModel):
    class Meta:
        proxy = True

MyModel.objects.create()
MyProxy.objects.create()

MyModel.objects.count() == 2
MyProxy.objects.count() == 1
MyProxy.objects.non_polymorphic().count() == 2
```

I think this is pretty confusing default behavior because the main point of a proxy model is to provide a different python interface to an existing db row and this by default both treats the proxies as polymorphic extensions and restricts the querysets to rows created from the proxy instance. There is certainly a case for it, and Django allows the ContentType system to opt into proxy distinctions (thought not by default) - I just think it would have made more sense for the default behavior to be what people naturally should expect.

Where to go from here:

1. I'm going to create [a discussion question](https://github.com/jazzband/django-polymorphic/discussions/689) to continue debate about what to do for version 5 and if any breaking changes are warranted to the default behavior.
2. I'm accepting this PR with some edits. It will allow users to toggle on more canonical behavior.

The new interface - which is now documented - will be:

```python
class MyModel(PolymorphicModel):
    ...

class MyProxy(MyModel):
    class Meta:
        # this removes the instance_of from the default polymorphic queryset and
        # also toggles proxy on
        polymorphic_proxy = True  

MyModel.objects.create()
MyProxy.objects.create()

MyModel.objects.count() == 2
MyProxy.objects.count() == 2
MyProxy.objects.instance_of(MyProxy).count() == 1
MyProxy.objects.non_polymorphic().count() == 2
```

The other change that I'm going to add to @pgammans PR is for proxy models to behave like proxies in querysets as well. For example, from above:

```python
for proxy in MyProxy.objects.all():
    assert isinstance(proxy, MyProxy)
```

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,10 +1,14 @@
 Changelog
 =========
 
+v5.0.0 (202x-XX-XX)
+-------------------
+
+* Implemented `Fix proxy model support. <https://github.com/jazzband/django-polymorphic/pull/676>`_
+
 v4.3.0 (202X-XX-XX)
 -------------------
 
-* Implemented `Add polymorphic_proxy flag to enable polymorphic queries from proxy models. <https://github.com/jazzband/django-polymorphic/pull/676>`_
 * Fixed `Resolve primary key name correctly. <https://github.com/jazzband/django-polymorphic/pull/620>`_
 * Implemented `Include get_child_inlines() hook in stacked inline admin forms. <https://github.com/jazzband/django-polymorphic/pull/681>`_
 * Fixed `multi-database support in inheritance accessors. <https://github.com/jazzband/django-polymorphic/pull/550>`_
--- a/src/polymorphic/tests/test_orm.py
+++ b/src/polymorphic/tests/test_orm.py
@@ -895,6 +895,47 @@ class PolymorphicTests(TransactionTestCase):
             self.assertEqual(2, NonAliasNonProxyChild.objects.count())
             self.assertEqual(3, NonProxyChild.objects.count())
 
+    def test_proxied_instances_are_proxies(self):
+        ProxyBase.objects.create(some_data="ProxyBase")
+        AliasProxyChild.objects.create(some_data="AliasProxyChild")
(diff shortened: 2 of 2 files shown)


bckohan:
On further consideration I'm bumping this PR to version 5 because I'm unsatisfied with any backwards compatible interface.

I do not believe that by default concrete model querysets should return proxy instances. I think that this behavior should be toggleable on by using the polymorphic meta flag on proxy instances, but should be off by default. That is different enough than the original intent of this PR that we have to bump it to 5.x

PR #681, opened 2025-12-08
Include get_child_inlines() hook in stacked inline admin forms.

### LINKED ISSUE #576: Add `PolymorphicInlineModelAdmin.get_child_inlines()`

bckohan:
Replaces #582 

Fixes #576

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -221,3 +221,4 @@ __marimo__/
 
 test1.db
 test2.db
+example/example.db
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -36,6 +36,7 @@
 * Jacob Rief
 * James Murty
 * Jedediah Smith (proxy models support)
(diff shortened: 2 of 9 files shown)


piranna:
The class hierarchy walks as default fool proof default behavior are the main reason of my previous PR #582 that you've closed, since this one would surpass It. If you are not willing to have that default behavior and instead raise an exception if child_classes is not defined (that IMHO would make this class useless being used alone without inheriting It in another child one), then please reopen again my PR.

PR #684, opened 2025-12-08
Add create_from_super method and test

### LINKED ISSUE #296: Create a sub table row from existing parent.
Hello, 

Would it be possible to extend an already saved polymorphic model?
```python
class A(PolymorphicModel)
    pass

class B(A):
    pass
```

would it be possible to turn an instance of `A` to an instance of `B` ?
like some sort of `B(A=A())`.
(I guess it would be technically possible since inheritance only store a pointer to the base class?)

bckohan:
This PR replaces #65 

It needs some more work. I would like it to support casting across multiple levels of hierarchy.

The discussion in #65 and the details of this [long standing Django ticket](https://code.djangoproject.com/ticket/7623) should be considered.

Fixes #296


bckohan:
It might be best to solve this upstream in Django - a solution [to this](https://code.djangoproject.com/ticket/7623) would also work for polymorphic models because they are just multi-table models with a fancy manager.

PR #741, opened 2025-12-18
fix: remove get_preserved_filters and add AdminPreservedFiltersTests

### LINKED ISSUE #125: Admin change form doesn't preserve changelist filter
When you save a polymorphic model form in the admin, you lose the changelist filter that was defined in your URL. This is because in `django/contrib/admin/templatetags/admin_urls.py`, Django compares the changelist url with the current one:

```python
def add_preserved_filters(...):
    # ...
    current_url = '%s:%s' % (match.app_name, match.url_name)
    changelist_url = 'admin:%s_%s_changelist' % (opts.app_label, opts.model_name)

    if changelist_url == current_url and '_changelist_filters' in preserved_filters:
        # Get preserved filters
```

Since the `changelist_url` is constructed based on `opts.model_name`, you get the "implementation" model name, not the base one, which is the one from `current_url`.  For example if you have a base model named "BaseTree" and an implementation named "OakTree", `changelist_url` would be `admin:tree_oaktree_changelist` but `current_url` would be `admin:tree_basetree_changelist`. So the 2 URLs don't match and the changelist filters are not correctly processed.

### LINKED ISSUE #356: Filters are not preserved in polymorphic parent admin
Inheriting from PolymorphicParentModelAdmin class break the django admin "preserve filter" behavior. Removing the "get_preserved_filters" in polymorphic/admin/parentadmin.py solves the problem.
Why is this override needed?

niltonpimentel02:
## Context

- I was able to reproduce the bug, and after removing the legacy code, the filters are working correctly again.
- It took me a while to understand how the admin tests are structured since I’m not very familiar with them.
- Once I figured that out, I added a test to ensure the behavior is now correct.
- I also tested it manually by creating a local project and installing the package with the updated code.

## Additional Comments

If I’m missing anything or overlooking some edge case, please let me know. I'm also open to any ideas or tips regarding the code!

Thanks, everyone 😄

## Related Issue

Closes #356
Closes #125

### DIFF
--- a/src/polymorphic/tests/test_admin.py
+++ b/src/polymorphic/tests/test_admin.py
@@ -647,24 +647,33 @@ class AdminRecentActionsTests(_GenericAdminFormTest):
 
 class AdminPreservedFiltersTests(_GenericAdminFormTest):
     def test_changelist_filter_persists_after_edit(self):
+        """
+        Test that changelist filters are preserved after editing an object.
+
+        Regression test for:
+        - #356: Filters are not preserved in polymorphic parent admin
+        - #125: Admin change form doesn't preserve changelist filter
+        """
         # Arrange: create 1 instance for each concrete polymorphic child model
         # so the changelist has something to filter and something to click into.
-        Model2B.objects.create(field1="B1", field2="B2")
-        Model2C.objects.create(field1="C1", field2="C2", field3="C3")
+        obj_b = Model2B.objects.create(field1="B1", field2="B2")
+        obj_c = Model2C.objects.create(field1="C1", field2="C2", field3="C3")
 
-        # Get the ContentType for Model2B. The admin filter uses this PK to
-        # restrict the polymorphic changelist to only "B" rows.
+        # Get the ContentType for Model2B to verify the filter later.
         ct_b = ContentType.objects.get_for_model(Model2B)
 
-        # Build a changelist URL for the polymorphic parent (Model2A) with the
-        # polymorphic content-type filter applied via querystring.
-        filtered_url = f"{self.list_url(Model2A)}?polymorphic_ctype__id__exact={ct_b.pk}"
+        # Act: Navigate to the changelist and apply the polymorphic content type filter
+        # by clicking the filter link in the admin UI.
+        self.page.goto(self.list_url(Model2A))
 
-        # Act: open the filtered changelist page in the browser.
-        self.page.goto(filtered_url)
+        # Click the filter link for Model2B in the polymorphic child model filter sidebar.
+        # Clicking a filter link navigates to a new URL, so we wait for navigation.
+        with self.page.expect_navigation(timeout=10000):
+            self.page.click("text=model2b")
 
         # Click the first row's object link in the results table to go to its change form.
-        self.page.click("table#result_list tbody tr th a")
+        with self.page.expect_navigation(timeout=10000):
+            self.page.click("table#result_list tbody tr th a")
 
         # Edit a field on the change form.
         self.page.fill("input[name='field1']", "B1-edited")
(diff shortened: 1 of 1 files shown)


src/polymorphic/tests/test_admin.py
```diff
@@ -521,3 +521,42 @@ def test_admin_recent_actions(self):
...
+        # Get the ContentType for Model2B. The admin filter uses this PK to
+        # restrict the polymorphic changelist to only "B" rows.
+        ct_b = ContentType.objects.get_for_model(Model2B)
+
+        # Build a changelist URL for the polymorphic parent (Model2A) with the
+        # polymorphic content-type filter applied via querystring.
+        filtered_url = f"{self.list_url(Model2A)}?polymorphic_ctype__id__exact={ct_b.pk}"
+
+        # Act: open the filtered changelist page in the browser.
+        self.page.goto(filtered_url)
```
bckohan: Instead of constructing the filtered url directly could we use playwright to click the filter button on the page?

src/polymorphic/admin/parentadmin.py
```diff
@@ -234,17 +234,6 @@ def delete_view(self, request, object_id, extra_context=None):
...
-        if "_changelist_filters" in request.GET:
-            request.GET = request.GET.copy()
-            filters = request.GET.get("_changelist_filters")
-            f = filters.split("&")
-            for x in f:
-                c = x.split("=")
-                request.GET[c[0]] = c[1]
-            del request.GET["_changelist_filters"]
-        return super().get_preserved_filters(request)
-
```
bckohan: Do we know why this code was added in #121? I feel like we're missing something.

src/polymorphic/tests/test_admin.py
```diff
@@ -521,3 +521,42 @@ def test_admin_recent_actions(self):
...
+            self.page.click("input[name='_save']")
+        response = nav_info.value
+
+        # Assert: request succeeded (admin returned a normal page load).
+        expected_status_code = 200
+        assert response.status == expected_status_code
+
+        # Assert: after saving, the redirected URL still contains the original filter,
+        # meaning the changelist preserved the querystring across edit/save.
+        assert f"polymorphic_ctype__id__exact={ct_b.pk}" in self.page.url
```
bckohan: Could we also verify that the change list is filtered by pulling the object ids off the page and comparing them to what we expect? I think the base admin test class has a get_object_ids function you can use.

bckohan:
Thanks for this @niltonpimentel02!

I added a few comments - This issue also relates to #125. I'd like this PR to close that one as well. My initial attempt at reproducing #125 wasn't successful. They're claiming a redirect after a save or update is redirecting from a url that has the derived class model type in it to a url that has the base class model type in it. I couldn't figure out how to navigate the admin pages to trigger that... but I do suspect this might partly be why the ``get_preserved_filters`` method was overriden in #121

niltonpimentel02:
> Thanks for this @niltonpimentel02!
> 
> I added a few comments - This issue also relates to #125. I'd like this PR to close that one as well. My initial attempt at reproducing #125 wasn't successful. They're claiming a redirect after a save or update is redirecting from a url that has the derived class model type in it to a url that has the base class model type in it. I couldn't figure out how to navigate the admin pages to trigger that... but I do suspect this might partly be why the `get_preserved_filters` method was overriden in #121

Hey! I forgot to mention that for my tests here, I installed `Django==4.2.27` and confirmed that removing the code, the filters worked as expected.

It would be nice to close both issues, but if we can't reproduce the error from issue 125, it's going to be tricky to resolve, as I managed to reproduce issue 356 here.

bckohan (approved):
thanks @niltonpimentel02! I did verify that this also fixes #125

PR #748, opened 2025-12-20
Docs: Add warning about orphaned rows from manual deletion

### LINKED ISSUE #352: orphaned polymorphic_ctype_id rows
While debugging with django-polymorphic i deleted some data without django-polymorhpic. The real instances table rows are gone, but in the base polymorphic table, the rows with `polymorphic_ctype_id` still exist.

BaseClass.objects.all() => empty Queryset
BaseClass.objects.non_polymorphic().all() => Complete Queryset with all "old" data.

if `polymorphic_ctype_id` does not match any real row i would expect an exception.

### 1.
how to delete the non_polymorphic-only BaseClass objects?
With non_polymorphic() i get existing BaseClasses plus "corrupt" BaseClasses. How to determine if a non_polymorphic object has no concrete object?

### 2.
how to avoid this corrupt state in the future? This should normally not happen, but i would feel better if i know the database never can enter this corrupt state.

JohananOppongAmoateng:
This PR updates the documentation to warn users about the consequences of manually deleting rows from child model tables (e.g., using raw SQL `DELETE` statements) without using Django's ORM.

Manual deletion bypasses Django's cascade deletion logic, leaving "orphaned" rows in the base model table. These rows retain a `polymorphic_ctype_id` pointing to the now-non-existent child, causing `BaseClass.objects.all()` to exclude them (correctly, as they are broken) while `BaseClass.objects.non_polymorphic().all()` still includes them.

Closes #352.


JohananOppongAmoateng:
> Great thankyou!!
> 
> I added a test to track this behavior. I suspect it will change in version 5.

Nice

PR #750, opened 2025-12-21
Fix: Admin Bugs

### LINKED ISSUE #380: Polymorphic inline form ignoring multipart for FileFields from Django 2.1
Admin form with polymorphic inlines is not multipart when you use an ImageField or a FileField and the files are not uploaded to the server.

This issue happens when implementing the example from documentation with StackedPolymorphicInline for django admin and having FileField or ImageField in the polymorphic models.

[Polymorphic Inline documentation](https://django-polymorphic.readthedocs.io/en/stable/admin.html#inline-models)

Using the identical example on versions of django prior to 2.1, works ok.

### LINKED ISSUE #612: Admin: add view popup breaks if initial submit has validation error
Scenario:
- in Django admin click green + next to the foreign key field that opens `PolymorphicParentModelAdmin` add view
- Submit form with validation error
- Fix the error and submit again

Expected:
- object is added, popup is closed, object is populated to the foreign key field widget on the original form

Actual:
- popup is not closed, field widget is not populated

This happens because after initial submit with validation error the query params indicating popup are removed, and only `ct_id` param is preserved. It comes from the following lines of code: https://github.com/jazzband/django-polymorphic/blob/master/polymorphic/admin/parentadmin.py#L202-L208

JohananOppongAmoateng:
### Issues Addressed

1. **Polymorphic inline forms ignoring multipart for FileFields**   #380  
2. **Admin add view popup breaks after initial validation error**   #612  

Closes  #380 
Closes  #612


bckohan (approved):
@JohananOppongAmoateng It looks like the change to helpers.py was unnecessary because #380 appears to have been fixed upstream in Django. I removed it and added a full integration test for uploads in inline forms. **Let me know if you think this is in error!**

PR #752, opened 2025-12-21
Implement best effort polymorphic object retrieval

### LINKED ISSUE #347: Getting a specific element from a queryset in an post_delete returns None
I encountered the following bug when using

- Django 2.0.4
- django-polymorphic 2.0.2
- Python 3.6.4
- Postgresql and sqllite

# Bug description

Create two instances of Classes that have the same Base class, using `django-polymorphic`. 
The Class of the first (!) instance has a post_delete signal

In this post_delete signal, you issue a Query to the database to determine what is the 'first' element of all the existing instances of the base class.

Now, delete the first instance, the one with the signal. 

If you do it like this, it does not work, because the [0] always returns `None`. If should however show the other instance.

```python
try:
    print(Entry.objects.order_by('pk')[0])
except IndexError:
    pass
```

if you do it like this, it also does not work, because the first() also always returns `None`. (first() is an easier form of [0] with the try/except)

```python
print(Entry.objects.order_by('pk').first())
```

If you do it like this, it does work!

```python
entry = None
for entry in Entry.objects.order_by('pk'):
    print(entry)
    break
```

Note that in all three case, when you issue

```python
print(Entry.objects.order_by('pk'))
```

You do see the Queryset, with the element that was not deleted.

### Notes

- The order in which the instances are created does matter (if your create the item with the signal secondly, the bug does not occur)
- The problem also occurs when using 'last()'

# Example code

I created a repo with tests that trigger the bug like mentioned above: https://github.com/megasnort/django_polymorphic_bug_first_last_post_delete

```
pip install -r requirements.txt`
cd django_first_bug
python manage.py test
```
You will see that two tests fail, that should pass. They don't because the first() and [0] method of the queryset return None instead of the first element.

JohananOppongAmoateng:
Fixes #347

## Problem
When objects were deleted in Django's post_delete  signals, calling first()  or __getitem__[0] on a polymorphic queryset would return None or raise IndexError, even when other objects existed in the queryset.
This occurred because:  Django deletes child rows before parent rows during multi-table inheritance deletion
When querying in a  post_delete  signal, parent rows still exist but child rows are gone. The old implementation filtered out objects where the derived class couldn't be fetched. This caused  first()  and __getitem__[0] to skip valid parent objects

Example
```
obj1 = Model2B.objects.create(field1="First", field2="B")
obj2 = Model2A.objects.create(field1="Second")
def signal_handler(sender, instance, **kwargs):
    # This returned None in the old implementation
    first_obj = Model2A.objects.order_by("pk").first()
    print(first_obj)  # Should be obj1 (as Model2A), but was None
post_delete.connect(signal_handler, sender=Model2B)
obj1.delete()
```
## Solution
Implemented a "best effort" approach that walks up the inheritance hierarchy when derived objects cannot be fetched:

When a derived object's row doesn't exist, try fetching it as each parent class.   Return the best available parent object instead of filtering it out. This matches Django's standard multi-table inheritance behavior


bckohan (changes requested):
Thanks for a solid start on this @JohananOppongAmoateng !

This is a tough one. There is a bug here but I would prefer not to override ``first()`` or ``__getitem__`` and instead fix the underlying issue in the iterator. The question is what should the behavior actually be? My hope is that [improvements in version 5](https://github.com/jazzband/django-polymorphic/discussions/743) might remove the need for tracking ``polymorphic_ctype_id`` all together and naturally eliminate issues like this one and #352 which is related. The iterator should *never* return None, which means we have two choices:

A. Remove the model object instead of returning None
B. Walk back up the inheritance hierarchy and return the correct model type. (I would refer to this as "best effort").

Your solution is A but I think we have to either document this behavior as a restriction - or implement B, both because B is more inline with what Django does with non-polymorphic models and because my hope is that this will be the behavior going forward in version 5.

I have moved the signal tests over into a test_signals module and have rewritten them to expect best effort (B). I have also added a test that demonstrates plain django behavior.

During deletion Django builds a graph of operations it must employ to accomplish the full deletion. These operations depend on each other and must happen in the correct order. Child rows must be deleted before parent rows and the signals are sent out at each step of this process. Therefore, when you query DB state in the middle of this operation you will get back undeleted parent rows. This is correct and expected if not a little obtuse.

bckohan:
I had slated this issue for the deletion fix release (4.5) but because it has high potentiality for regressions I'm going to kick it to the ORM fixes release (4.6)

JohananOppongAmoateng:
Actually my first thought was to return the parent model but i wasn't sure whether that was supposed to be the expected behavior. But i do agree with you it is a though one probably something we should slate for 4.6 or v5

JohananOppongAmoateng:
@bckohan i have addressed your concerns

bckohan:
@JohananOppongAmoateng please hold off on any work on this. I'm working on the following adjustments:

1. Use _meta.parents to walk up the concrete model hierarchy
2. Batch retries to avoid N+1 queries for failed retrievals

bckohan:
Basically done, but blocked pending fix for #781

bckohan:
The solution implemented here orders the queries per each model class in ``_get_real_instances`` based on the class position in the inheritance tree - leaves first. This allows us to retry failed retrievals of non-existent rows higher up the hierarchy. As a last resort the base object will be returned. The iterator should never return null. This solution also batches the retry queries needing only one extra per class in the inheritance chain rather than one per failed row retrieval.

Issue #472, opened 2020-11-27
Override multiple different sections of base_fieldsets


Lorac:
I have a model with extra fields but I would like to show these extra fields inside already existing fieldsets from the form base class.

I was able to do it with something like

```
def merge_by_key(ts):

    dic = {}
    for i, j in ts:
        dic.setdefault(i,{}).setdefault('fields', ())
        dic[i]['fields'] += tuple(j['fields'])

    return tuple((k, v) for k, v in dic.items())
```

I can only override one section here because it's based on the `extra_fieldset_title`

This doesn't control where the fields end up in the fieldsets.

If I use `extra_fieldset_title` with the same name, well I just get 2 fieldsets with the same names.

For a case like this, should I just fully override base_fieldsets for each subclasses? Which is kind of annoying.

PR #756, opened 2025-12-23
FormSet Bugs

### LINKED ISSUE #363: Nested polymorphic_inline_formsets gives AttributeError: 'NoneType' object has no attribute 'get_real_instance_class'
I am trying to create a nested inline formset with the innermost being the polymorphic inline formset

#this my polymorphic inline formset
Oformset=polymorphic_inlineformset_factory(MBF, MO, 
                                           formset_children=(PolymorphicFormSetChild(LO),
                                                             PolymorphicFormSetChild(DO),
                                                             PolymorphicFormSetChild(BO),
                                                             PolymorphicFormSetChild(AO),
                                                             ),fields='__all__', can_delete=False, extra=0)

#I created a regular baseinlineformset and added the Oformse, polymorphic_inlineformset, to it
class MyBaseFormset(BaseInlineFormSet):
    def add_fields(self, form, index):
        super(BaseScenarioFormset, self).add_fields(form, index)
        form.nested=Oformset(instance=form.instance,
                                  data=form.data if form.is_bound else None,
                                  files=form.files if form.is_bound else None,
                                  prefix='modelobject-%s-%s' % (form.prefix,
                                                                Oformset.get_default_prefix())
                                  )    
    
    def is_valid(self):
        result =super(BaseScenarioFormset, self).is_valid()     
        if self.is_bound:
            for form in self.forms:
                result=result and form.nested.is_valid()
        return result
    
    def save(self, commit=True):
        result=super(BaseScenarioFormset, self).save(commit=commit)
        for form in self:
            form.nested.save(commit=commit)
        return result

ModelBasicFormSet=inlineformset_factory(UCP, MBF, fields=['modelNum','interactor', 'action', 'validationStep'],
                                            formset=MyBaseFormset, can_delete=False,extra=0)

### LINKED ISSUE #549: Issue with polymorphic_ctype when populating polymorphic inline formsets.
When trying to populate a polymorphic inline formset using the django-polymorphic package, I'm encountering a problem with the `polymorphic_ctype` field.

**Context**

In the `get_context_data` method, I populate the formset's initial data using:

```python
ct = ContentType.objects.get_for_model(constraint)
initial[i]['polymorphic_ctype'] = ct
```
This works for an initial GET request to display the form. However, upon a POST, the formset becomes populated with the POST data where `polymorphic_ctype` is a string (eg "scheduler | sun separation constraint") and not the expected id, leading to an error:

["Formset row constraint_set-0 has no 'polymorphic_ctype' defined!"]

**Expected Behavior**

The `polymorphic_ctype` should be correctly recognized.

**Current Behavior**

[The code](https://github.com/django-polymorphic/django-polymorphic/blob/master/polymorphic/formsets/models.py#L205) expects an instance of ContentType:

```python
model = defaults["initial"]["polymorphic_ctype"].model_class()
```

However, [it](https://github.com/django-polymorphic/django-polymorphic/blob/master/polymorphic/formsets/models.py#L189) also tries to retrieve an id from the POST data, which causes the inconsistency:

```python
ct_id = int(self.data[f"{prefix}-polymorphic_ctype"])
```

Django version: 4.2.4
django-polymorphic version: 3.1.0
Python version: 3.11.5

Is there a better method to populate the formset? I'm attempting this approach to implement a "Duplication" feature. The intention is for the form to display data from the original Job and its associated constraints. This data has not been saved yet, allowing the user to make any necessary corrections.

```python
 class JobDuplicateView(JobCreateView):
    def get_context_data(self, **kwargs):
        data = super().get_context_data(**kwargs)

        cloned_constraints = self.duplicated_constraints
        
        # Prepopulate the formset with the cloned constraints
        initial = [{} for _ in cloned_constraints]
        for i, constraint in enumerate(cloned_constraints):
            constraint_form = data['constraint_formset'].get_form_class(constraint.__class__)()

            ct = ContentType.objects.get_for_model(constraint)
            initial[i]['polymorphic_ctype'] = ct

            for name, field in constraint_form.fields.items():
                if name != "id" and name != "job":
                    initial[i][name] = getattr(constraint, name)

        data['constraint_formset'].initial = initial
        data['constraint_formset'].extra = len(cloned_constraints)

        return data
```

### LINKED ISSUE #578: PolymorphicFormSetChild overrides form exclude
Hi!

I have these models:

```python
from django.db import models
from polymorphic.models import PolymorphicModel

class Author(models.Model):
    pass

class Book(PolymorphicModel):
    author = models.ForeignKey(Author, on_delete=models.CASCADE)

class SpecialBook(Book):
    pass
```

I want to make a formset to edit special books of a given author:

```python
from django import forms
from polymorphic.formsets import polymorphic_modelformset_factory, PolymorphicFormSetChild

class SpecialBookForm(forms.ModelForm):
    class Meta:
        exclude = ("author",)

SpecialBookFormSet = polymorphic_modelformset_factory(Book, formset_children=(
    PolymorphicFormSetChild(SpecialBook, form=SpecialBookForm),
))
```

By doing so, the field `author` is not excluded. Indeed, due to [this line](https://github.com/jazzband/django-polymorphic/blob/fa6bc9f10b45fbe950e8db8af41fa3b35380d129/polymorphic/formsets/models.py#L51), `exclude` is not `None` when passed to `modelform_factory()` [here](https://github.com/jazzband/django-polymorphic/blob/fa6bc9f10b45fbe950e8db8af41fa3b35380d129/polymorphic/formsets/models.py#L98) then `form.Meta.exclude` is overridden [here](https://github.com/django/django/blob/main/django/forms/models.py#L618).

What's the point of `self.exclude = exclude or ()` [here](https://github.com/jazzband/django-polymorphic/blob/fa6bc9f10b45fbe950e8db8af41fa3b35380d129/polymorphic/formsets/models.py#L51)?

Thanks!

JohananOppongAmoateng:
This PR addresses multiple long-standing edge cases in django-polymorphic related to formsets and admin customization. Specifically, it improves how polymorphic formsets handle child form configuration, polymorphic_ctype initialization, and fieldset extensibility in subclassed models.

The changes aim to make polymorphic forms more predictable, composable, and aligned with standard Django behavior, while removing the need for fragile workarounds in downstream projects.

Closes #578 
closes #549 
Closes #363


bckohan:
@JohananOppongAmoateng I dropped the fix for #472 because that use case is too much scope creep for my comfort. I converted the issue to a discussion and answered it. So as not to lose your code I moved it to a [branch here](https://github.com/jazzband/django-polymorphic/tree/base_fieldset_merge).

bckohan:
I also removed the docs about m2m inline forms and directed that issue to #808 instead.

Issue #505, opened 2021-12-05
can the instance_of Q-object be made to accept a string?


johncronan:
I have found that, in my base model class, I want to add a UniqueTogether to the constraints in the `Meta`, with a condition of the form:

```~Q(instance_of=DerivedModel)```

where `DerivedModel` is inheriting from this base class. Because of the inheritance, I have to define DerivedModel _after_ this one, but that also means it's not yet defined for me to reference in the Meta class.

Can this Q-object be made to accept a string, as well? That's how other components, such as ForeignKey, deal with the same type of problem with circular dependency.

Or is there a solution here that I'm not thinking of?

AdamDonna:
This is probably an issue that needs to be resolved in Django rather than in django-polymorphic

johncronan:
The main reason I might want to have it that deep would be to use the constraint for serializing on natural keys.

In the case I'm dealing with, there's another combination of fields that works for that. So this is not a big deal.

But I don't see how Django could resolve it, since instance_of is implemented here. In _get_mro_content_type_ids, we query the Content Type framework with get_for_model. The alternative is get_by_natural_key:

https://docs.djangoproject.com/en/4.0/ref/contrib/contenttypes/#django.contrib.contenttypes.models.ContentTypeManager.get_by_natural_key

> The primary purpose of this method is to allow ContentType objects to be referenced via a [natural key](https://docs.djangoproject.com/en/4.0/topics/serialization/#topics-serialization-natural-keys) during deserialization.

bckohan:
I think theres a bug in github because when I merged #757 #766 was also merged. I'll just review 766 and if it needs any changes we'll issue another PR.

JohananOppongAmoateng:
Making a new pr and doing an interactive rebase to try and remove that commit

PR #757, opened 2025-12-23
ORM Bugs

### LINKED ISSUE #446: Content types, multiple databases and database routers
First of all, great package! I've been using it for years.

I have a multi database setup. I need the content types created on all databases. In order to determine which models are used on each database, I am entirely relying in a database router. When creating a polymorphic model, the instance is correctly getting the database for write from the router, however, the content type is ignoring this as the referenced code makes it an explicit choice that must be passed. In the absence of an explicit choice, and because it's a creation, the content types are always read from the default database. Since Django doesn't allow relations between databases, I get an error when trying to save the instance.

https://github.com/django-polymorphic/django-polymorphic/blob/eed4e9fa4a8c69e158fadbcec03fc153cb059f35/polymorphic/models.py#L73-L85

Is there any way to overcome this situation without having to explicitly pass a database to my `polymorphic_instance.save(...)` ?

Thanks in advance.

### LINKED ISSUE #486: ForeignKeyViolation when trying to `save` an entity when using a non-default db
I am facing the issue that I have two databases with different data in `django_content_type` tables. I am trying to copy entities from one db into the other and running into the problem that the `polymorphic_ctype` is not updated properly and causing a `ForeignKeyViolation`.

```python
entity = PolymorphicModel.objects.using("from_database").first()
entity.save("to_database")
```

`polymorphic_ctype` will be set on the entity with the id from `from_database`. When trying to save the entity to the `to_database` the id of the content_type might not be defined in the `to_database` or might be linked to a different model. IIUC `pre_save_polymorphic` is intended to update the `polymorphic_ctype` to the correct content type of the target database. In my case this is not working as the `polymorphic_ctype` is already defined.

My current workaround is to update the `polymorphic_ctype` using very similar code before saving:

```python
entity = PolymorphicModel.objects.using("from_database").first()
new_c_type = ContentType.objects.db_manager(using="to_database").get_for_model(entity.get_real_instance_class())
entity.polymorphic_ctype_id = new_c_type.id
entity.save("to_database")
```

Would it make sense to have a `force_update_c_type` flag for `pre_save_polymorphic/save` to enforce the update of the `polymorphic_ctype`?

```python
# Suggested usage
entity = PolymorphicModel.objects.using("from_database").first()
entity.save("to_database", force_update_c_type=True)
```

JohananOppongAmoateng:
This PR addresses multiple issues to improve django-polymorphic's reliability and usability.

Saving polymorphic objects from one database to another caused ForeignKeyViolation because polymorphic_ctype_id referenced ContentType IDs from the source database.

Database routers weren't respected - ContentType was always fetched from default database even when routers specified a different database.

Closes #486 
Closes #446


JohananOppongAmoateng:
> Great work @JohananOppongAmoateng
> 
> I'd like to keep PRs logically homogenous. Could you please separate out the fix for #505 into a separate PR? It is unrelated to the db routing issues (which I want to accept as is). I'm not sure about supporting 505 yet or not.

@bckohan  your requested changes have been done

bckohan (approved):
Great thanks!

PR #765, opened 2025-12-27
Add raw ID field filtering for polymorphic models

### LINKED ISSUE #144: Can `django-polymorphic` filter the popup change list for raw ID fields?
I have several polymorphic child models with M2M fields to each other, NOT to the parent model. I'm using the raw ID widget for these fields in the admin, but the popup change list shows a list of ALL child types of child model.

It seems that `django-polymorphic` would be in the best position to provide a customised `ForeignKeyRawIdWidget` that automatically filters the parent model's change list by content type.

Is this something you would be interested in merging?

Can you provide any guidance on how it might be implemented in `django-polymorphic`?

I am imagining the raw ID widget would just add `?polymorphic_ctype=N` to the URL for the popup link IF the foreign key pointed to a child model, and `PolymorphicParentModelAdmin.get_queryset()` would filter the queryset on that if set.

JohananOppongAmoateng:
This PR implements automatic filtering of popup change lists for raw ID fields when they point to polymorphic models.

Closes #144

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -113,6 +113,83 @@ Child model types can be filtered by adding a
 :attr:`~django.contrib.admin.ModelAdmin.list_filter` attribute. See the example above.
 
 
+Using Raw ID Fields with Polymorphic Models
+--------------------------------------------
+
+When using :attr:`~django.contrib.admin.ModelAdmin.raw_id_fields` for foreign keys or many-to-many
+fields that point to polymorphic models, the popup change list normally shows **all** child types.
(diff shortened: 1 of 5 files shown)


bckohan:
I'm closing the related issue as won't fix. This convenience function is either too specific to the user or has been fixed. I have tested the scenario where child models A and B have M2M fields to each other and the raw ids correctly only show A or Bs respectively. The user may have put a self referential M2M relationship in the base class and in that case showing ids for all child model types is correct behavior.

PR #766, opened 2025-12-29
can the instance_of Q-object be made to accept a string

### LINKED ISSUE #505: can the instance_of Q-object be made to accept a string?

JohananOppongAmoateng:
Q-objects couldn't accept string model references, causing circular dependency issues in Meta constraints when a base model needed to reference a derived model defined later.

Closes #505

### DIFF
--- a/src/polymorphic/query_translate.py
+++ b/src/polymorphic/query_translate.py
@@ -114,9 +114,11 @@ def _translate_polymorphic_filter_definition(
     # handle instance_of expressions or alternatively,
     # if this is a normal Django filter expression, return None
     if field_path == "instance_of":
-        return create_instanceof_q(field_val, using=using)
+        return create_instanceof_q(field_val, using=using, queryset_model=queryset_model)
     elif field_path == "not_instance_of":
-        return create_instanceof_q(field_val, not_instance_of=True, using=using)
+        return create_instanceof_q(
+            field_val, not_instance_of=True, using=using, queryset_model=queryset_model
+        )
     elif "___" not in field_path:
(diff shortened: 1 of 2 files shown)


bckohan:
Somehow when I merged #757 this one was merged too?

JohananOppongAmoateng:
Oops that's weird shouldn't have happened

JohananOppongAmoateng:
i dont know if i should do an interactive rebase on master

JohananOppongAmoateng:
@bckohan do you think that could solve this?

bckohan:
@JohananOppongAmoateng So looking at master the commit from this PR was *not* merged after all. I think github's internal bookkeeping had registered this commit on the other PR so when it saw it closed it just marked this PR closed. We can't reopen this PR so I think the thing to do here is just open a new PR from that same branch and cross our fingers!

JohananOppongAmoateng:
mm okay will try that and lets see

bckohan:
@JohananOppongAmoateng when you get a chance could you please open a new PR from this branch?

PR #779, opened 2025-12-31
feat: Enhance pre-commit hooks to match CI validation

rodrigobnogueira:
Improves the pre-commit configuration to provide local feedback that matches CI, reducing the need for multiple PR iterations due to linting issues.

### Pre-commit Configuration (.pre-commit-config.yaml)

**New Hook Structure:**
- **Auto-fix hooks** (run first on changed files):
  - `lint`: Auto-fixes import sorting and linting issues via ruff
  - `format`: Auto-fixes code formatting via ruff

- **Validation hooks** (match CI exactly):
  - `check-lint-strict`: Mirrors `just check-lint` from CI
  - `check-format-strict`: Mirrors `just check-format` from CI
  - `check-package`: Mirrors `just check-package` (dependency check)
  - `check-readme`: Mirrors `just check-readme` (README rendering)

- **Manual hooks:**
  - `run-tests`: Available for manual test execution

**Key Improvements:**
- Direct `uv run ruff` commands instead of `just` delegation
- `pass_filenames: true` for faster incremental checks
- `types: [python]` filtering for efficiency
- `verbose: true` for clearer development feedback

### Ruff Configuration (pyproject.toml)

- Added `force-exclude = true` to ensure exclusions work with `pass_filenames: true`
- Added `[tool.ruff.format].exclude` for migration file exclusions

## Why This Matters

Before this change, developers would push code, wait for CI, discover lint failures, fix them, push again, and repeat. This created a slow feedback loop.

Now, the pre-commit hooks:
1. Auto-fix common issues before commit
2. Validate that code will pass CI checks locally
3. Properly exclude migration files
4. Provide verbose, helpful output

## Testing

All hooks pass on the full project:
```
Lint (auto-fix)...............Passed
Format (auto-fix).............Passed
Check Lint (CI match).........Passed
Check Format (CI match).......Passed
Check Package (CI match)......Passed
Check README (CI match).......Passed
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,18 +1,70 @@
 ci:
   # Don't run these in pre-commit.ci at all
-  skip: [lint, format]
+  skip: [lint, format, check-lint-strict, check-format-strict, check-package, check-readme, run-tests]
 
 repos:
   - repo: local
     hooks:
+      # Auto-fixing hooks (run first) - runs on changed files only
       - id: lint
-        name: Lint
-        entry: just lint
+        name: Lint (auto-fix)
+        entry: bash -c 'uv run ruff check --fix --select I "$@" && uv run ruff check --fix "$@"' --
         language: system
-        pass_filenames: false
+        pass_filenames: true
+        types: [python]
+        stages: [pre-commit]
+        verbose: true
 
       - id: format
-        name: Format
-        entry: just format
+        name: Format (auto-fix)
+        entry: uv run ruff format
+        language: system
+        pass_filenames: true
+        types: [python]
+        stages: [pre-commit]
+        verbose: true
+
+      # Validation hooks - runs on changed files only
+      - id: check-lint-strict
+        name: Check Lint (CI match)
+        entry: bash -c 'uv run ruff check --select I "$@" && uv run ruff check "$@"' --
+        language: system
+        pass_filenames: true
+        types: [python]
+        stages: [pre-commit]
+        verbose: true
+
+      - id: check-format-strict
+        name: Check Format (CI match)
+        entry: uv run ruff format --check
+        language: system
+        pass_filenames: true
+        types: [python]
+        stages: [pre-commit]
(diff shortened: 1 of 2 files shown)


JohananOppongAmoateng:
i dont know if its a good idea and would be great but i think it would be nice to add running specific test files that change for each commit.

rodrigobnogueira:
Thanks for the feedback!

**@copilot suggestions:**
-  Removed redundant `**/migrations/*.py` pattern
- Simplified `check-readme` to use `uv run -m readme_renderer`
- Kept `uv pip check` instead of `uv run pip check` - the latter fails because `pip` is a uv subcommand, not a Python script. Note: this means the justfile's `check-package` task may also have this issue.

@JohananOppongAmoateng  Great idea about running specific tests for changed files. Never done that before.

JohananOppongAmoateng:
> Thanks for the feedback!
> 
> **@copilot suggestions:**
> 
> * Removed redundant `**/migrations/*.py` pattern
> * Simplified `check-readme` to use `uv run -m readme_renderer`
> * Kept `uv pip check` instead of `uv run pip check` - the latter fails because `pip` is a uv subcommand, not a Python script. Note: this means the justfile's `check-package` task may also have this issue.
> 
> @JohananOppongAmoateng Great idea about running specific tests for changed files. Never done that before.

Just something that came into my mind i dont know if its even possible but i was thinking if we make a change in file a we most likely would add new tests or remove tests for that file so i expect to see some tests files changing that could be run just for those. I am mean it just an idea dont know how feasible it is and whether its worth it

bckohan:
> i dont know if its a good idea and would be great but i think it would be nice to add running specific test files that change for each commit.

This would be nice but I'd like to keep commits fast and low barrier so its easy for people to share code. There are many cases where you might want to commit a failing test (WIP PRs or to demonstrate an issue in a branch).

bckohan:
Thanks for the effort on this, some of the ideas here got rolled into #856

PR #788, opened 2026-01-02
Allow forward references in instance_of queries.

### LINKED ISSUE #505: can the instance_of Q-object be made to accept a string?

JohananOppongAmoateng:
Closes #505

### DIFF
--- a/src/polymorphic/query_translate.py
+++ b/src/polymorphic/query_translate.py
@@ -114,9 +114,11 @@ def _translate_polymorphic_filter_definition(
     # handle instance_of expressions or alternatively,
     # if this is a normal Django filter expression, return None
     if field_path == "instance_of":
-        return create_instanceof_q(field_val, using=using)
+        return create_instanceof_q(field_val, using=using, queryset_model=queryset_model)
     elif field_path == "not_instance_of":
-        return create_instanceof_q(field_val, not_instance_of=True, using=using)
+        return create_instanceof_q(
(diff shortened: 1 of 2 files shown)


bckohan:
@JohananOppongAmoateng can you think of any real use cases that a forward reference would enable? I think the OPs desire in #505 is to put a forward reference in the condition parameter of a Constraint. Unfortunately this solution won't work for them because no polymorphic manager is involved in that scenario.

bckohan:
I'm bumping this to 4.10+ so we can have some time to conjur up a reasonable use case. It seemed simple enough to do but if theres no real need for it then it'd just be more code to maintain.

JohananOppongAmoateng:
> @JohananOppongAmoateng can you think of any real use cases that a forward reference would enable? I think the OPs desire in #505 is to put a forward reference in the condition parameter of a Constraint. Unfortunately this solution won't work for them because no polymorphic manager is involved in that scenario. 

Cant think of anything

PR #792, opened 2026-01-03
Document post_save signal behavior during loaddata (#502)

### LINKED ISSUE #502: Instance attributes missing in post save signal triggered by loaddata
Hello,

I'm a bit lost. I have created a function to create a celery beat task on a post save signal for child objects of a polymorphic class. When the signal is triggered while django is running, all is working fine. However, when the signal is triggered while loading fixtures (manage.py loaddata), two weird issues arise:

1/ The id field and the pk field of the instance doesn't match
2/ field inherited from the parent polymorphic class aren't set (the name field)

Models:

```
class Endpoint(PolymorphicModel):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    name = models.CharField(max_length=250, verbose_name="Name")

class GenericSwitch(Endpoint):
    pass

class EXSwitch(GenericSwitch):
    def healthcheck(self):
        pass
```

```
@receiver(post_save)
def endpoint_saved_event(sender, instance, created, **kwargs):
    if created and hasattr(instance, 'healthcheck'):
        try:
            logger.info(msg={
                "message": f"Paramétrage automatique de la vérification de l'état de santé de {instance.name} ({instance.__class__}/{instance.id}/{instance.pk})"
            })
```

When triggered via loaddata, the name attribute is not retrieved and the pk/id fields do not match.

```
[2021-10-20 11:28:34][core.signals][INFO][None] ::: {'endpoint': {'name': '', 'id': 'f96e3d6e-dc93-4907-aab3-a257536c1282'}, 'message': "Paramétrage automatique de la vérification de l'état de santé de  (<class 'core.models.EXSwitch'>/f96e3d6e-dc93-4907-aab3-a257536c1282/dc04c0ba-a4bb-4704-a440-83329994caac)"}
```

When triggerd via regular instance create, the name attribute is retrieved and the pk/id fields match.

```
[2021-10-20 11:44:52][core.signals][INFO][e164139de3aa46a2b65ef34e77cb2acd] ::: {'endpoint': {'name': 'Stack test', 'id': 'ef2a01b3-9f76-49fd-b71d-4d523a483320'}, 'message': "Paramétrage automatique de la vérification de l'état de santé de Stack test (<class 'core.models.EXSwitch'>/ef2a01b3-9f76-49fd-b71d-4d523a483320/ef2a01b3-9f76-49fd-b71d-4d523a483320)"}
```

Any idea why polymorphic acts differently ?

Thanks.

JohananOppongAmoateng:
Users reported that when using loaddata to load fixtures for polymorphic models, post_save signal handlers receive instances with:

Mismatched pk and id fields
Missing parent class attributes (empty values)
This only occurs during loaddata, not during normal instance creation.

Root Cause
This is expected Django behavior for multi-table inheritance during deserialization, not a bug in django-polymorphic.  Closes #502

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -136,16 +136,23 @@ Working with Signals and Fixtures
 ----------------------------------
 
 When using Django's :django-admin:`loaddata` command with polymorphic models, you may notice that
-``post_save`` signal handlers receive instances that appear incomplete - parent class attributes
+``post_save`` signal handlers receive instances that appear incomplete – parent class attributes
 may be empty and ``pk``/``id`` fields may not match. **This is expected Django behavior** for
 multi-table inheritance during deserialization, not a bug in :pypi:`django-polymorphic`.
 
 Understanding the Issue
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-During fixture loading, Django deserializes parent and child table rows separately. When a child
-model's ``post_save`` signal fires, Django passes a ``raw=True`` parameter to indicate the data
-is being loaded from a fixture. At this point, parent attributes may not yet be fully accessible.
+During fixture loading, Django deserializes parent and child table rows separately
+(see :Django:`fixture loading <topics/db/fixtures/>`).
+
+When a child model's ``post_save`` signal fires, Django passes a ``raw=True`` parameter to indicate
+the data is being loaded from a fixture. At this point, parent attributes may not yet be fully
+accessible.
+
+Django’s :Django:`post_save <ref/signals/#post-save>` signal is called with ``raw=True`` during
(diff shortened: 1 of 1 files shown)


bckohan (changes requested):
This is looking good thanks @JohananOppongAmoateng 

I assume this is discussed in the django documentation? Can we add a link to it using intersphinx references? There's a shortcut for pulling down the intersphinx objects using just:

```bash
just fetch-refs django
```

JohananOppongAmoateng:
@bckohan added the refs

bckohan:
On further consideration this is just normal Django behavior and I don't think it's our responsibility to document. The issue was converted to Q&A and answered.

PR #826, opened 2026-01-11
Fix repeated migrations in Django 5.x by relaxing PolymorphicGuard equality

### LINKED ISSUE #825: 4.9 generates a lot of migrations
This was an issue that happened in 4.5  and it seems to happen again in 4.9. See  #759

JohananOppongAmoateng:
## Description
Fixes an issue where `django-polymorphic` generates repeated/redundant migrations in Django 5.x (and potentially 4.x) because [PolymorphicGuard](cci:2://file:///c:/Users/DELL/Documents/GitHub/django-polymorphic/src/polymorphic/deletion.py:33:0-98:39) wrappers around [on_delete](cci:1://file:///c:/Users/DELL/Documents/GitHub/django-polymorphic/src/polymorphic/tests/test_migrations/test_on_delete.py:299:4-344:85) handlers were not comparing equal to the original handlers (like `models.CASCADE`) during migration state detection.

The fix involves updating `PolymorphicGuard.__eq__` to allow direct equality comparison with the wrapped action function.

Closes #825

### DIFF
--- a/src/polymorphic/deletion.py
+++ b/src/polymorphic/deletion.py
@@ -71,6 +71,8 @@ class PolymorphicGuard:
         return migration_fingerprint(self.action)
 
     def __eq__(self, other):
+        if self.action == other:
+            return True
         if (
             isinstance(other, tuple)
             and len(other) == 3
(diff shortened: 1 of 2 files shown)


JohananOppongAmoateng:
Closing this

PR #848, opened 2026-01-15
Fix infinite recursion bug with use of only and certain field access patterns.

### LINKED ISSUE #844: bug: Infinite recursion with django-filer and django-polymorphic>=4.9
I am running into an issue with django-filer's test suite as soon as I update django-polymorphic to version 4.9 or later: It seems that using the `Q` object leads to an infinite recursion.

```
...
  File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/django/db/models/sql/where.py", line 218, in clone
    clone = self.create(connector=self.connector, negated=self.negated)
  File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/django/utils/tree.py", line 36, in create
    obj = Node(children, connector or cls.default, negated)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

I have not tracked the bug down yet, and it might well be part of django-filer, but I wanted to put this here, in case anyone else has a similar problems. 

I'll keep this issue updated once I know more.

bckohan:
closes #844 

(reproduces the issue, working on a fix)


bckohan:
I'm not sure this is fixable given the two-stage iterator process. This may have to be bumped to version 5.

PR #856, opened 2026-01-22
feat: add comprehensive type annotations

### LINKED ISSUE #647: Add type hints for django-polymorphic
Hi @bckohan,

It looks like our project currently doesn’t have type hints, so tools like mypy or pyright can’t properly perform static type checking. I’d love to contribute by adding type hints to the project.

There are two ways we could approach this:
	1.	Add type hints directly within this project – This would involve modifying a large portion of the codebase.
	2.	Contribute external type hints to the [typeshed](https://github.com/python/typeshed) project – This keeps the main codebase untouched. I’ve done something similar before for the django-channels package ([PR #13939](https://github.com/python/typeshed/pull/13939)). The downside is that types in typeshed would need to be kept in sync with this project over time.

I believe adding type hints will improve maintainability and tooling support for the project. I’m happy to follow whichever approach you prefer, and I’d appreciate your help reviewing the code once I get started.

What do you think about this suggestion?

huynguyengl99:
Close #647

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ sphinx = true
 
 [tool.ruff]
 line-length = 99
+target-version = "py310"
 
 exclude = [
     "**/migrations/*.py",
@@ -209,7 +210,7 @@ typing = [
     "mypy>=1.19.1",
 ]
 lint = [
-    "ruff>=0.9.8", 
+    "ruff>=0.15.0", 
     "readme-renderer[md]>=43.0",
     "doc8>=1.1.2"
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -761,7 +761,7 @@ dev = [
     { name = "pytest-mock", specifier = ">=3.15.1" },
     { name = "pytest-playwright", specifier = ">=0.7.2" },
     { name = "readme-renderer", extras = ["md"], specifier = ">=43.0" },
-    { name = "ruff", specifier = ">=0.9.8" },
+    { name = "ruff", specifier = ">=0.15.0" },
     { name = "sphinx", specifier = ">=7.1.2" },
     { name = "sphinx-autobuild", specifier = ">=2024.10.3" },
     { name = "sphinxcontrib-django", specifier = ">=2.5" },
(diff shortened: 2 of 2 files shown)


huynguyengl99:
Hi @bckohan and @JohananOppongAmoateng, this PR is based on the type annotations I added in the `typeshed` repo. Could you take a look at it? Also, if everything looks good, could you approve the GitHub workflows to run, @bckohan?

huynguyengl99:
Hi @bckohan, could you re-approve the workflow run? I’ve updated the tests, so I expect it should pass now.

There are still quite a few typing errors, but I think we can address them in future commits in this PR—either when I have more time, or with the help of @JohananOppongAmoateng.

huynguyengl99:
Hi @bckohan , could you re-approve the workflow run?

huynguyengl99:
Hi @bckohan, cc @JohananOppongAmoateng 

I’ve finished addressing all remaining typing issues in the recent commits on this branch. Here’s a brief summary of the current state and some context around the decisions made:

**Summary**
- Due to the complexity of the project and the need to support both mypy and pyright, the current implementation requires:
  + 3 # pyright: ignore comments
  + 63 # type: ignore comments
  + 47 explicit cast(...) calls
- I made a strong effort to minimize the use of ignore comments and casts. The remaining ones are, to the best of my judgment, the most compact and reasonable solutions given the constraints.

**Rationale**
- One major reason for these workarounds is that we use inline type hints, while projects like Django and Django REST Framework rely on separate type stub packages.
- Inline typing makes certain cases harder, but I believe this approach will help us long-term and avoids the need to constantly sync a separate typeshed-style repo with the main codebase.
- This PR is quite large, and I’ve spent a significant amount of time on it. I’d really appreciate it if you could review it when possible to catch any incorrect or suboptimal type hints early — that will help avoid larger conflicts later.

**Supporting both typed and non-typed users**

I’ve tried to ensure compatibility for both typed and non-typed users. For example:
```python
_ModelT = TypeVar(
    "_ModelT",
    bound=PolymorphicModel,
    default=PolymorphicModel,
)

if TYPE_CHECKING:
    _ModelAdminBase = admin.ModelAdmin[_ModelT]
else:
    _ModelAdminBase = admin.ModelAdmin

class PolymorphicChildModelAdmin(_ModelAdminBase, Generic[_ModelT]):
    ...
```
- I haven’t tested this pattern extensively yet, but the test suite is currently passing, so it should be working as expected. 

**On cast(...) vs type: ignore**
- Where mypy complains, I generally prefer using cast(...) instead of # type: ignore. For example:
```python
return cast(type[BaseGenericPolymorphicInlineFormSet], FormSet)
```
- While this may add a bit of inconvenience during development, I think explicit casts are clearer and better communicate intent — especially when dealing with Django’s heavy metaprogramming.
- Unfortunately, I don’t see a cleaner alternative in these cases, so I hope the casts are reasonable and acceptable.

Looking forward to your feedback. Thanks a lot for taking the time to review this — I hope we can get this PR merged soon.

PR #907, opened 2026-07-09
fix: `KeyError` when accessing fields deferred by base manager

### LINKED ISSUE #906: Fields deferred by base manager raise `KeyError` when accessed.
I was in the process of upgrading from an ancient version of django-polymorphic (4.0.0) and noticed a regression that isn't covered by tests.

The problem I found relates to #334. That change injects `"polymorphic_ctype_id"` into the `fields` argument of `refresh_from_db`. Following the call chain, `fields` gets passed to  `super().only()` in PolymorphicQuerySet. 

The unintended side-effect of that is to interfere with how Django tracks which fields are deferred.

Django uses [a set difference](https://github.com/django/django/blob/b34b834378adc590a771cccd44b9fb0b953cdca2/django/db/models/sql/query.py#L2545) to determine which deferred fields have been passed to `.only()`, and tracks those via `Queryset.deferred_fields`. For example, if there is 1 deferred field in the queryset, and we pass that field to `.only()`, Django will determine that that field needs to be loaded. (In this case, that all fields should be read.)

With `polymorphic_ctype_id` being injected into fields, that set difference doesn't detect when a deferred field has been passed to `.only()` correctly.

I should note that I have only been able to replicate this behaviour where the base manager for a polymorphic model overrides `get_queryset` in order to defer one or more fields. This is the established pattern in the codebase I am working with, where you have a field containing large data (e.g. JSON blobs) which do not need to be read in the majority of cases.

This manifests itself primarily as a `KeyError` raised in Django's `DeferredAttribute` descriptor, [here](https://github.com/django/django/blob/b34b834378adc590a771cccd44b9fb0b953cdca2/django/db/models/query_utils.py#L282), when you try to access a field (which is deferred by a custom manager) on a model instance.

I have a fix for this that I intend to raise a PR for shortly.

chrisimcevoy:
Ensure that injecting `"polymorphic_ctype_id"` into `deferred_loading` happens after Django's `.only()` implementation has determined which deferred fields to read.

closes #906


src/polymorphic/tests/test_orm.py
```diff
@@ -1487,6 +1488,14 @@ def test_refresh_from_db_fields(self):
         obj.refresh_from_db(fields=["field1"])
         assert obj.field1 == "aa1"
 
+    def test_access_field2_deferred_by_manager(self):
+        Model2BDeferredManager.objects.create(field1="aa", field2="bb")
+
+        obj = Model2BDeferredManager.objects.get()
+
+        assert obj.field1 == "aa"
+        assert obj.field2 == "bb"
```
chrisimcevoy, line 1497: Without the fix, accessing `field2` here raises `KeyError`.

bckohan (approved):
This is a solid, correct fix. I traced the mechanics and verified it end-to-end.

Root cause. Django stores field-loading intent as query.deferred_loading = (field_set, defer_flag):
- (fields, True) → defer these (load everything else)
- (fields, False) → load only these
- (∅, False) → the empty-only special case, which Django's deferred_to_data treats as load all fields

The scenario: a model whose base_manager defers field2 (DeferredManager.get_queryset().defer("field2")). When you later access obj.field2, Django's DeferredAttribute.__get__ calls refresh_from_db(fields=["field2"]), which runs ._base_manager.only("field2").get(). That base-manager queryset already carries deferred_loading = ({"field2"}, True). Django's add_immediate_loading then computes {"field2"} - {"field2"} = ∅, flipping to (∅, False) — i.e. "load everything."

The old only() unconditionally injected polymorphic_ctype_id before calling super().only(), so the result became ({"polymorphic_ctype_id"}, False) — "load only ctype." field2 stayed deferred, the refresh never fetched it, and re-reading data["field2"] blew up with KeyError: 'field2'.

The fix moves the injection to after super().only() and gates it on the resulting only-set being non-empty. In the collapse-to-load-all case it leaves (∅, False) alone (all fields load, including ctype); in the normal .only("field1") case it re-adds ctype exactly as before.

